PR #1, opened 2018-10-18
Bids report

Remi-Gau:
Here is the adapted version of the bids_report funtion from the BIDS starter kit.

One part of the function still has a call to spm_vol if we want try to get data from the header.


gllmflndn:
Thanks @Remi-Gau, I just pushed your changes. We can then work on the SPM dependency in another issue.

PR #8, opened 2019-07-14
small fix in a filter that skipped json files when they were in the root folder

Remi-Gau:
The filter had one underscore that can disappear when the json file is put in the root folder of the data set. 

Not thoroughy tested (unit-test required ?).

Might be worth updating in `spm_BIDS.m` too.


gllmflndn:
Thanks @Remi-Gau, could you link to a dataset and a query that illustrate the issue? An alternative fix could be to use `_?` as a reminder that we would expect a `_` most of the time.

Adding unit testing is definitely required - with continuous integration through Octave. Could you open another issue for that (and have your test function moved there)?

Remi-Gau:
Hard to find an open dataset out there that that shows the behavior so I am modifying the MoAE BIDS dataset to illustrate.

Can you confirm you get this behavior too?

```
% download typical BIDS example
urlwrite('https://www.fil.ion.ucl.ac.uk/spm/download/data/MoAEpilot/MoAEpilot.bids.zip','MoAEpilot.zip');
unzip(fullfile(pwd,'MoAEpilot.zip'));

% create a T1w json file in the base directory
jsoncontent.FlipAngle = '5';
bids.util.jsonencode(fullfile(pwd,'MoAEpilot', 'T1w.json'), jsoncontent)

% try to get metadata
BIDS = bids.layout(fullfile(pwd,'MoAEpilot'));

% works for func but comes out empty for T1w
metadata = bids.query(BIDS, 'metadata', 'sub', '01', 'type', 'bold');
metadata = bids.query(BIDS, 'metadata', 'sub', '01', 'type', 'T1w')
```

**Expected**
metadata = 

  struct with fields:

    FlipAngle: '5'


**Actual**
metadata = 

  struct with no fields.

gllmflndn:
OK, thanks! I removed the unit test from this pull request to have it part of issue #9 instead.

Remi-Gau:
Sweet jesus. I am not even sure why I had commited that "test" file in here. Sorry for the confusion. Thanks for bearing with my inaptitudes.

Issue #3, opened 2018-11-13
Test on octave: dealing with classes ('+' subfolders)


Remi-Gau:
I am trying to get some stuff running on a pure octave + SPM12 environment so while I wanted to test this matlab-bids there to see if it behaved well with octave.

Tested on the empty dataset ds001 from the BIDS-example repo.

Running the following that relies on the SPM machinery (i.e `spm_BIDS.m`) works fine
```
docker run -it --rm \
-v /c/Users/Remi/Documents/BIDS/ds001/:/data: \
-v /c/Users/Remi/Documents/BIDS/code/:/code \
spmcentral/spm:octave-latest eval "spm_BIDS('/data/')"
```

But trying to do the same without using SPM by running this fails.
```
docker run -it --rm \
-v /c/Users/Remi/Documents/BIDS/ds001/:/data: \
-v /c/Users/Remi/Documents/BIDS/code/:/code \
spmcentral/spm:octave-latest script '/code/test_bids_matlab.m'
```

With `bids-matlab` in the `/code` folder  and with `test_bids_matlab.m` being:
```
addpath(genpath(fullfile('/code', 'bids-matlab')))
data_dir = '/data/';
BIDS_dir = fullfile(data_dir);
layout  = bids.layout(data_dir)
```

This is the error message I get.

```
Execution failed: /code/test_bids_matlab.m
'file_utils' undefined near line 24 column 16
Error in layout (line 24)
Error in test_bids_matlab (line 4)
Error in spm (line 1218)
Error in run (line 71)
Error in spm (line 1218)
Error in spm_standalone (line 115)
Error in /usr/local/bin/spm12 (line 38)
```

Looks like older version of octave did not deal well with subfolders to define classes (that's what "+" folders are, correct?).
http://octave.1599824.n4.nabble.com/Are-package-folder-supported-td4630147.html

Is that still the case with octave 4.4 (the one that this container uses)?

Am I missing some matlab / octave subtle difference here?

gllmflndn:
Hello,
This is a bug in Octave about private directories in +package:
https://savannah.gnu.org/bugs/?54658
https://savannah.gnu.org/bugs/?45444
I still went ahead and used that feature because I hope it will be fixed at some point and in the meantime one can add explicitly the `+bids` directory in the path - you will get a warning that `addpath: package directories should not be added to path` but the code will run.
This makes me notice a compatibility issue in file_utils.m; I push a short fix shortly.

[...]

PR #10, opened 2019-07-20
[WIP] Unit tests - read metadata

Remi-Gau:
This is a first unit test that adds 
- a dummy data set (with dummy files made with `touch`) and 
- 4 unit-tests to check metadata reading for T1w and bold and check the good implementation of the inheritance principle.

It seems that the way metadata are read currently has things "backwards": higher level overrides lower levels rather than the other way around as the [specs want it](https://github.com/bids-standard/bids-specification/blob/master/src/02-common-principles.md#the-inheritance-principle): _The values from the top level are inherited by all lower levels unless they are overridden by a file at the lower level._ 

With this dummy data set 

`bids.query(BIDS, 'metadata', 'type', 'bold');`

and 

`metadata = bids.query(BIDS, 'metadata', 'sub', '01', 'type', 'bold');`

should return different values (or am I misunderstanding something) as the content of `sub-01_task-auditory_bold.json` change a value from the higher levels.

But the 2 calls give the same thing. So this leads to a failed test when running for example the lines below.

```
%% test func metadata base directory
metadata = bids.query(BIDS, 'metadata', 'type', 'bold');
assert(metadata.RepetitionTime == func.RepetitionTime);

%% test func metadata subject 01
metadata = bids.query(BIDS, 'metadata', 'sub', '01', 'type', 'bold');
assert(metadata.RepetitionTime == func_sub_01.RepetitionTime);
```


Remi-Gau:
Currently trying to check it the tests run with octave but there is the issue of the "+" directories ( see #3 ) that is getting in the way. Will keep investigating but pointers are welcome. :-)

gllmflndn:
Ah, I merged too early and the issue you mentioned about metadata and the inheritance principle hasn't been solved. We should open a new issue or PR.

Remi-Gau:
OK will try to implement this then.

PR #11, opened 2019-07-23
Tolerant option for bids.layout

tanguyduval:


tanguyduval:
This issue solves #11 
Tolerance on:
- dataset_description and .tsv files
- parse image modalities that are not in BIDS yet (but in extension proposal for instance)

Other tolerant conditions that could be added:
- inconsistent data between subjects (currently it crashes if more images in one subject than another)

Remi-Gau:
Nice.
Wouldn't it sense to have the inconsistency between subjects be the lowest level of tolerance: unless I am mistaken this does not make an invalid bids data set, where as the other 2 levels would.
Or am I missing something?

tanguyduval:
@Remi-Gau agreed, I just tried an inconsistent dataset with bids validator and it gives no error. Only a warning in case of inconsistent data between subjects:
````
Warning 2: [Code 38] INCONSISTENT_SUBJECTS
Not all subjects contain the same files. Each subject should contain the same number of files with the same naming unless some files are known to be missing.
````

I'll change that.
Best

tanguyduval:
@Remi-Gau this version accepts inconsistent data between subjects, even for files not officially in BIDS yet
tested on octave 5 and matlab.
Best

Remi-Gau:
Thank. Will check things out during the weekend

Remi-Gau:
- [ ] add tests to check warning VS error handling of the tolerant option of layout

Remi-Gau:
If that's with everyone I will start mergin those old pull-requests that have not been "vetoed".
- bring the PR up to date with master
- clean up and refactor if needed
- create tests if needed or where easily possible
- wait for 24 hours after CI gives the green light (in case anyone has any final thoughts before merging)
- merge

Remi-Gau:
On this one we would need some test to assert that a warning or an exception is thrown. 

Not sure that our current framework allows this.

Remi-Gau (approved):
Good with me

PR #12, opened 2019-07-23
Query

tanguyduval:
I could not query the session list for a specific subject. This branch solves this issue.


gllmflndn:
Thanks! I guess I didn't think about querying for 'sessions' in that way. Would it be possible to back out e36e626 as it relates to PR #11 and add one unit test instead?

tanguyduval:
Done. I had to use the folder Synthetic instead of ds007 to test for sessions.
Cheers

gllmflndn:
Thank you!

PR #17, opened 2019-10-03
add support for scans.tsv when building the BIDS layout 

robertoostenveld:
The content of scans.tsv is added in the same manner as the subjects (which are technically subject/session combinations).

```
making layout for synthetic
>> BIDS
BIDS = 
  struct with fields:

             dir: '/Users/roboos/tmp/bids-examples/synthetic'
     description: [1×1 struct]
        sessions: {}
           scans: [10×1 struct]
            sess: [0×0 struct]
    participants: [1×1 struct]
        subjects: [10×1 struct]
```

I have not added any code for the querying yet, because that needs more discussion on the design.


robertoostenveld:
withdrawing PR due to lack of progress

Issue #16, opened 2019-10-02
when creating the layout, the user should be able to specify whether to read (metadata) or not


robertoostenveld:
Right now most of the metadata is automatically read (bval, bvec, events, channels), but other metadata is not read (headshape, photo, etc.). I think this mainly distinguishes between the metadata being in `tsv` files (which are read) versus another format (which are not read).

I think it would be nice to make this configurable, so that the user would know whether he can always expect a MATLAB struct or a string pointing to a filename.

gllmflndn:
I agree. I was not sure what to do in a first instance, that's why there are all these `% ?` in the code. Not only is this inconsistent but it might affect significantly the parsing time and the amount of memory required.
Adding an option in `BIDS.layout` makes sense. I will factor that in with pull request #11. Do you have in mind something like reading everything or nothing, or something more fine-grained?

robertoostenveld:
I am actually not sure yet... I will try it out on more datasets to see what feels right.

robertoostenveld:
Actually I think that it would be fine to keep it as it is, i.e. whenever it relates to a tsv or json file, that it returns the content.

PR #19, opened 2019-10-04
clarify the design and user interface

robertoostenveld:
This is a proposal to clarify the design and to serve to improve/complete the implementation.

It allows edits from maintainers, so please use the issue commenting and edit where appropriate. 

The writing style (as "proposal") is still not appropriate for the actual readme, so needs more work once we agree upon the design and API and user-interface.


README.md
```diff
@@ -31,6 +31,124 @@ A [tutorial](https://github.com/bids-standard/bids-matlab/blob/master/examples/t
...
+BIDS.sessions     % from sessions.tsv, array of Nsubjects x 1
+BIDS.subjects     % data for each modailty in each subject*sesssion
+BIDS.scans        % from scans.tsv, array of (Nsubjects*Nsessions) x 1
+```
+
+In the layout, `sessions` has (conceptually) the same size as `participants` (although differently represented).
+
+In the layout, `scans` has the same size as `subjects`.
+
+I propose to rename `subjects` to `subses`, since it corresponds to subjects _and_ sessions. Every `subses` has the modalities (anat, func, eeg, etc) as structure arrays, where each array corresponds to one data file. For example
```
robertoostenveld: this needs a decision

README.md
```diff
@@ -31,6 +31,124 @@ A [tutorial](https://github.com/bids-standard/bids-matlab/blob/master/examples/t
...
+BIDS.query(layout, 'sessions') returns an array of strings
+BIDS.query(layout, 'modalities') returns an array of strings, like {'eeg', 'anat'}
+BIDS.query(layout, 'types') returns an array of strings, like {'eeg', 'T1w', 'T2w'}
+BIDS.query(layout, 'tasks') returns an array of strings
+BIDS.query(layout, 'runs') returns an array of strings
+```
+
+For subjects, sessions and modalities it can come from the directory structure. For tasks and runs it has to come from the `key-value` entities in the file names. The same could also be used for subjects ('sub-xxx') and sessions ('ses-xxx'), which are in the file name, but the modalities are not as key-value pairs in the file name.
+
+I propose that when you specify
```
robertoostenveld: this needs a decision

README.md
```diff
@@ -31,6 +31,124 @@ A [tutorial](https://github.com/bids-standard/bids-matlab/blob/master/examples/t
...
+
+```
+BIDS.query(layout, 'data') returns an array of strings
+```
+
+All other queries of metadata should return an array that is the same size as the array of data files.
+
+#### Querying for different types of metadata
+
+The following returns the content of the JSON files that correspond to the data files (see above). In principle it could also return a list of strings with the file names of the JSON files.
```
robertoostenveld: whether to return an array of file names or an array with actual content is discussed in #16.

README.md
```diff
@@ -31,6 +31,124 @@ A [tutorial](https://github.com/bids-standard/bids-matlab/blob/master/examples/t
...
+
+```
+BIDS.query(layout, 'metadata') returns an array of structures
+```
+
+The following returns electrode positions for each data file (see above). In principle it could also return a list of strings with the file names of the TSV files.
```
robertoostenveld: whether to return an array of file names or an array with actual content is discussed in #16.

robertoostenveld:
@gllmflndn @arnodelorme @ftadel @CPernet @vlitvak, I would like your input on this. 

For those of you not yet aware of this repository: it is an effort to make a common MATLAB interface between (current and future) BIDS datasets and our MATLAB-based software packages. The idea is that this interface will be maintained to reflect updates to the BIDS specification, so that we don't have to do that multiple times ourselves. Also, this interface should deal with the complexities such as inheritance of metadata, and resolve the (known) ambiguities in a consistent manner.

ftadel:
Thanks, Robert.
Adding @martcous to the discussion, he's dealing with part of the BIDS I/O in Brainstorm.

arnodelorme:
Robert, do you think this should be merged with https://github.com/sccn/bids-matlab-tools. This repo contains tools to specifically deal with EEGLAB import and export to BIDS. However, there is also support files to export tsv and json files. 

Arno

+bids/layout.m
```diff
@@ -110,10 +110,10 @@
 for su=1:numel(sub)
     sess = cellstr(file_utils('List',fullfile(BIDS.dir,sub{su}),'dir','^ses-.*$'));    
     for se=1:numel(sess)
-        if isempty(BIDS.subjects)
-            BIDS.subjects = parse_subject(BIDS.dir, sub{su}, sess{se});
+        if isempty(BIDS.subses)
```
apjanke: What do you think about using `subjs` instead of `subses`? In Matlab, `subs` has a somewhat conventional meaning of "subscripts" (that is, a multidimensional indexing vector), and "subses" could be interpreted as "an array of subscript arrays".

README.md
```diff
@@ -31,6 +31,126 @@ A [tutorial](https://github.com/bids-standard/bids-matlab/blob/master/examples/t
...
+The field `participants` represents the participants.tsv details.
+
+The field `sessions` has (conceptually) the same size as `participants`, although differently represented. It represents the session.tsv details for each subject.
+
+The field `scans` has the same size as `subses`. It represents the scans.tsv details for each subject and session.
+
+The field `subses` represents the data files in all subjects _and_ optionally in all sessions. Every `subses` has the modalities (anat, func, eeg, etc) as structure arrays, where each array corresponds to one data file. For example
+
+```
+>> BIDS.aubses(1)
```
apjanke: Typo: should be ".subses" here?

README.md
```diff
@@ -31,6 +31,126 @@ A [tutorial](https://github.com/bids-standard/bids-matlab/blob/master/examples/t
...
+       ieeg: [0×0 struct]
+        pet: [0×0 struct]
+```
+
+### User interface
+
+Users are not meant to directly interface with the BIDS layout object/structure, but rather use the `query` function. In general that works like
+
+```
+BIDS.query(layout, 'item') returns a complete list of the specific items
```
apjanke: The use of `BIDS` here seems inconsistent from the `BIDS.description` etc. used above. There, `BIDS` is a variable holding the BIDS layout structure. Here, `BIDS` looks like it's referencing the `+bids` package, but using the wrong case. (This may be an issue on systems with case-sensitive filesystems, like Linux.) I suggest using the `BIDS` identifier in the same way everywhere, instead of switching between using `BIDS` above and `layout` here to refer to the object/structure returned from `bids.layout(...)`.

apjanke:
What do you all think about using an object instead of a struct for the `BIDS` structure (the thing that's returned from `bids.layout` and passed to `bids.query(BIDS,...)` and `bids.report(BIDS,...)`? When `query()` or `report()` receive an argument, it needs to be either a char (filename) or a BIDS structure in the proper format. To ensure that, it really ought to be validating that the BIDS argument is a struct with the correct fields, instead of some other random struct that the user created and passed in. Using objects would make this easier, because then all you'd have to test is whether it's an object of the right type using `isa(BIDS, '<class-name>`)`, instead of checking the set of fields and their values on a struct. An object could also be a convenient place to stick some of the parsing and querying logic, in ways that would make it easier to unit test, even if that interface wasn't exposed to end users and they were supposed to treat it as a custom struct. You could also do custom display formats, which could make it easier on users once the `BIDS` structure gets really big, with fields to support eeg and other extensions.

Matlab R2014a and Octave 4.4 both have decent support for MCOS objects, so I don't think compatibility is a concern for using Matlab/Octave's OOP support here.

+bids/layout.m
apjanke: Oh, wait: I missed the part about this now meaning "subjects *and* sessions". Sorry.

+bids/query.m
```diff
@@ -23,11 +23,16 @@
 opts = parse_query(varargin);
 
 switch query
+<<<<<<< HEAD
```
apjanke: Looks like you got a couple unresolved merge conflicts here.

robertoostenveld: bummer!  There should not have been any code changes in here; I suspect that I packed this on top of other changes.  

I will make a new PR with only changes to the README file.

robertoostenveld:
> What do you all think about using an object instead of a struct for the `BIDS` structure (the thing that's returned from `bids.layout` and passed to `bids.query(BIDS,...)` and `bids.report(BIDS,...)`?

Agreed, that would provide better encapsulation of the details.

robertoostenveld:
Question: The consequence of making a proper object out of it would also be that the functions/methods would look different, right? Like this?

```
dataset = bids('bids-examples/ds000117')
query(dataset, 'subjects')   % returns an array of strings like {'01', '02', ...}
query(dataset, 'sessions')   % returns an array of strings
query(dataset, 'modalities') % returns an array of strings, like {'eeg', 'anat'}
query(dataset, 'types')      % returns an array of strings, like {'eeg', 'T1w', 'T2w'}
query(dataset, 'tasks')      % returns an array of strings
query(dataset, 'runs')       % returns an array of strings
```

apjanke:
> The consequence of making a proper object out of it would also be that the functions/methods would look different, right? Like this?

That is entirely optional. You could change them to be like that (which is what pybids does, I think). Or you could just let all the fields on `dataset` be public and leave the logic in `bids.query()`, treating `dataset` as a glorified struct that just enforces its field structure. Or you could do a mix of both: implement some of the logic as smaller-size methods inside the `dataset` class, and leave the `bids.query()` function as the main interface for users to interact with it, leaving them the option to dig down into the `dataset` class's methods if they want to do advanced programming.

robertoostenveld:
objects are also interesting to deal with the probably soon-to-be-merged [derivatives](https://github.com/bids-standard/bids-specification/pull/265), which in themselves are BIDS datasets. E.g. for this https://openneuro.org/datasets/ds000117/versions/1.0.3 (which is more recent than the bids-examples version) it would be

```
dataset = bids('bids-examples/ds000117')
query(dataset, 'derivatives')   % returns an array of BIDS datasets
````

which would return an array similar to 
```
array{1} = bids('bids-examples/ds000117/derivatives/freesurfer')
array{2} = bids('bids-examples/ds000117/derivatives/meg_derivatives')
```

apjanke:
You could also entirely mimic the existing PyBIDS interface/class design. If it's a more mature library, that could mean their interface is good. And it might make it easier for users to switch between the two languages, if that's a concern.

apjanke:
I put together another draft PR that's an example of how you could go OOP while maintaining compatibility with the current interface: #39.

robertoostenveld:
this is outdated and not suitable for merging. I do hope that in one of the coming meetings we will be able to discuss how to move toward a API design specification.

Remi-Gau:
Yes I think that this would be next in line right after we have a better idea of what we exactly need from the code base.

This should happen early because changing this later is going to be a world of pain.

PR #21, opened 2019-10-21
Add some extra comments to better explain each function

Remi-Gau:


gllmflndn:
Thanks @Remi-Gau: I made a few changes to use the imperative and capitalise the first letter in comments. I also added a plaster so that the tests pass (unrelated issue).

PR #24, opened 2019-12-20
util: Support jsonencode/jsondecode as regular functions

apjanke:
What do you think about supporting jsonencode/jsondecode when implemented as regular functions, in addition to just when they're built-ins? This would allow the jsonencode/jsondecode from jsonstuff (https://github.com/apjanke/octave-jsonstuff) or any other Matlab-compatible implementation to be used, in addition to JSONio. This would be nice for me because I'd rather use jsonstuff than JSONio. And there's a chance that jsonencode/jsondecode will eventually get merged into Octave core from one of these libraries, but might be regular functions instead of builtins there, so this would be forward-looking for future Octave versions. (Also, Matlab could always change the jsonencode/jsondecode implementation to use regular M-file wrapper functions, which would break the current detection.)

### DIFF
--- a/+bids/+util/jsondecode.m
+++ b/+bids/+util/jsondecode.m
@@ -12,9 +12,15 @@ function value = jsondecode(file, varargin)
 % Copyright (C) 2018, Guillaume Flandin, Wellcome Centre for Human Neuroimaging
 % Copyright (C) 2018--, BIDS-MATLAB developers
 
+persistent has_jsondecode
+if isempty(has_jsondecode)
+    has_jsondecode = ...
+        exist('jsondecode','builtin') == 5 || ...       % MATLAB >= R2016b
+        ismember(exist('jsondecode','file'), [2 3]);    % jsonstuff or other Matlab-compatible implementation
+end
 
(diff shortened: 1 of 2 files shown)


gllmflndn:
Yes. JSONio is only here temporarily until we can assume Matlab is recent enough to have `jsonencode`/`jsondecode` and an agreement is reached in Octave:
https://savannah.gnu.org/bugs/?53100
Shouldn't the test with `exist` also look for an M-file (ie not only `3`)? It would also perhaps be better if the checks were only done once with the outcome stored in a persistent variable (as a function handle).
Having `jsonstuff` pushed into Octave before the 6.1 release would be great. If there isn't a quick agreement of which C++ library to rely on, `jsonencode` could still be added as I guess everyone implemented it in pure M-code.

apjanke:
You're right; it should be looking for both 3 and 5. I have amended this PR to do so, and to store the outcome in a persistent variable. (Though maybe you effectively wanted the entire if/elseif/.../end chain logic and all those tests stored in a persistent.)

PR #25, opened 2019-12-20
Move +bids/private stuff to +bids/+internal?

apjanke:
What do you think of moving the package-private functions in `+bids/private` to a new `+bids/+internal` package?

This would allow these functions to be directly tested and played with during development, but would still indicate to users of the library that they're for BIDS-MATLAB's internal use and shouldn't be called from client code. This "internal" package name is a pretty well-established convention these days. I've seen it in both Java and Matlab code, including MathWorks' own code.

This also removes the need to do the extra `addpath` step under Octave to work around its handling of `private` directories inside packages. This PR updates the `.travis.yml` file to reflect that.

### DIFF
--- a/+bids/+internal/README.md
+++ b/+bids/+internal/README.md
@@ -1,0 +1,6 @@
+BIDS-MATLAB +internal package
+=============================
+
+This package contains functions for BIDS-MATLAB's internal use only.
+
+Do not call these functions directly from your code! They are undocumented and subject to change at any time. If you call these functions from your code, your code may break or produce incorrect results!
--- a/+bids/+internal/file_utils.m
+++ b/+bids/+internal/file_utils.m
@@ -1,11 +1,11 @@
 function varargout = file_utils(str,varargin)
 % Character array (or cell array of strings) handling facility
-% FORMAT str = file_utils(str,option)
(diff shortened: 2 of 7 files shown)


robertoostenveld:
it is indeed used a lot in MATLAB toolboxes. I am fine with this.

Remi-Gau:
I will defer to people with more experience on this (AKA @robertoostenveld) so I would be in favor.

gllmflndn:
The issue in Octave with `private` directories inside +packages has now been [fixed](https://savannah.gnu.org/bugs/?54658) and will be available in the next release.

Changing `private` to `+internal` is a switch from a hard constraint to a soft one on who can call the function. I initially preferred `private` as the rule is then explicit and enforced but it is true that `+internal` will allow to write more fine-grained unit tests and will allow these functions to be called throughout the `+bids` package (e.g. from `+bids/fcn1.m` and `+bids/+util/fcn2.m` so that's a :+1: for me too.

Remi-Gau (approved):

Remi-Gau:
Will merge this one as well in 24 hours.

Speak now or ~~forever~~... send a PR later.

Issue #22, opened 2019-12-09
Supported Matlab versions?


apjanke:
Hi, BIDS folks,

Could you document which versions of Matlab are supported by bids-matlab?

And to what degree is Octave supported? Is bids-matlab strictly limiting itself to features & functions available in Octave as well as Matlab? And if so, which versions of Octave are supported?

gllmflndn:
Hi @apjanke,

The original code which `bids-matlab` is based upon was written to work with all MATLAB versions since R2007a and 'recent' Octave (4.2.2 onwards).

This is now to be decided for `bids-matlab` so I hope others will chime in. I would argue R2014a for MATLAB. For Octave we are limited by the version that users easily have access to with their package manager. Still, we can be forward-looking and make sure that we will implement in Octave missing features that will be used in this project.

robertoostenveld:
MATLAB R2014a and up would work for me (i.e. for FieldTrip). I don't know enough about Octave versioning to give an informed opinion about that.

robertoostenveld:
Related to this: Last week I was at http://practicalmeeg2019.org where we did hands-on sessions with ~40 people attending the session I was chairing. R2014 (a or b) was the oldest one that I encountered there, some from 2015, and most 2017-2018.

apjanke:
> MATLAB R2014a and up would work for me (i.e. for FieldTrip).

Sounds good to me.

> I don't know enough about Octave versioning to give an informed opinion about that.

I work with Octave. I'd recommend Octave 4.2.2 onwards like you're doing now. 4.2.x is the latest version available in many Linux package managers. So it's in common use. And 4.4.1 is the latest Octave version that works stably on macOS.

PR #27, opened 2019-12-20
Fix or suppress M-lint code inspection warnings

apjanke:
Just a couple M-lint warning fixes. With this PR merged, bids-matlab will be M-lint clean. (In Matlab R2014b, anyway. In R2019b, there are some new M-lint warnings that pop up. But you can't be clean in both versions: if you suppress the new R2019b warnings, then R2014b will give you warnings about spurious inspection suppressions. You can't win. ;) )

### DIFF
--- a/+bids/+util/tsvread.m
+++ b/+bids/+util/tsvread.m
@@ -130,7 +130,7 @@ n2  = cellfun(@(x) strcmpi(x,'NaN'),var);
 if header && any(n1 & ~n2)
     hdr     = true;
     try
-        var = genvarname(var);
+        var = genvarname(var); %#ok<DEPGENAM>
     catch
         var = matlab.lang.makeValidName(var,'ReplacementStyle','hex');
         var = matlab.lang.makeUniqueStrings(var);
--- a/+bids/private/file_utils.m
+++ b/+bids/private/file_utils.m
@@ -17,6 +17,7 @@ function varargout = file_utils(str,varargin)
 
 % Copyright (C) 2011-2018 Guillaume Flandin, Wellcome Centre for Human Neuroimaging
 
+%#ok<*AGROW>
 
 if ismember(lower(str), {'list','fplist'})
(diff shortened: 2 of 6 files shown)


gllmflndn:
I am personally not obsessed with some of the M-lint warnings and finds that following them makes the code uglier but it might be a matter of taste. The use of `~` and `genvarname` is/was here for compatibility with old Matlab versions (same with `catch err` in other pull requests).

I am not available in the next couple of weeks to review any changes but @robertoostenveld has admin rights and you and @Remi-Gau have write permissions in this repository.

apjanke:
Okay. If you'd rather not do the warning suppressions, I can just close this out.

The `[~] = ...` syntax was introduced [in Matlab R2009b](https://www.mathworks.com/help/matlab/release-notes.html?rntext=tilde&startrelease=R2009b&endrelease=R2017a&groupby=release&sortby=descending&searchHighlight=tilde), so it's surely supported by everything BIDS-MATLAB is targeting now. (Over [here](https://github.com/bids-standard/bids-matlab/issues/22) we talked about supporting R2014a+.)

Similarly, the "`catch err`" syntax was introduced [in Matlab 7.5 and `lasterror` became "Not recommended" in R2009b](https://www.mathworks.com/help/matlab/release-notes.html?rntext=catch&startrelease=R2008b&endrelease=R2017a&groupby=release&sortby=descending&searchHighlight=tilde), so it should be safe to use now.

robertoostenveld:
I agree to make the code consistent with Mathworks coding recommendations for our oldest supported version (i.e. 2014a)

Remi-Gau:
Sort of digging this one out from its inactivity. :-)

I like keeping the almighty linter god happy so I would be in favor of merging this. 

After a rebase for an update.

@apjanke I will take care of that.

I will also use some of the comment we have made here to start describing what matlab / octave versions we intend to support. I have the feeling we are going to need it if we want more contribution from the community. :-)

PR #28, opened 2019-12-20
Use semicolons in if/else/end one-liners

apjanke:
Matlab's code style recommends using semicolons after the "else" when you're writing a one-liner if/else/end statement like these. These show up as M-lint inspection warnings in newer versions of Matlab.

This is just code formatting cleanup; no behavior changes.


apjanke:
Oops, never mind about this: turns out Matlab changed its mind somewhere between R2014b and R2019b; adding these semicolons causes M-lint warnings to appear in R2014b. Canceling this PR since bids-matlab supports versions back to R2014a, and I suspect you'll want to prioritize earlier versions?

PR #30, opened 2019-12-21
More detailed error messages

apjanke:
Here's a suggestion for more detailed error messages. This mostly involves including the actual problematic values in error messages, including a specific reason/underlying error message for a failure, or telling which BIDS file/dir it was trying to process when the error occurred. This will make it easier for users to diagnose errors when using BIDS-MATLAB.

It also switches to using the `catch <err>` syntax where appropriate. This syntax is supported by all Matlab and Octave versions indicated in https://github.com/bids-standard/bids-matlab/issues/22, and is more concise than calling `lasterror()`.


apjanke:
:+1:

PR #32, opened 2019-12-21
Make function helptext more concise with a +bids/Contents.m

apjanke:
What do you think about making the per-function helptext more concise by concentrating the whole-library doco in a central `+bids/Contents.m` help file (that pops up when you do `help bids`)?

This way users can do `help bids` to get info on the overall library, and `help <fcn>` for the individual functions takes up less precious space in the command window. Each of the individual function help texts will have a link to the main `bids` help item so it'll be easy for users to find.

Works in Matlab R2014b-R2019b, though there's a bug in earlier versions of Matlab that causes the help for `help bids` to be displayed twice.

### DIFF
--- a/+bids/+util/tsvread.m
+++ b/+bids/+util/tsvread.m
@@ -1,5 +1,5 @@
 function x = tsvread(f,v,hdr)
-% Load text and numeric data from file
+% Load text and numeric data from tab-separated-value or other file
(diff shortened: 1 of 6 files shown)


robertoostenveld:
good idea!

apjanke:
Thanks for all the merges!

PR #33, opened 2019-12-21
util.tsvread: Make it a recursive call to self

apjanke:
There's no plain tsvread() function in Matlab; this looks like it was an attempt by tsvread to call itself on the contents of an unpacked .gz file. To do this, `tsvread` needs to be fully-qualified by its package, even though it's calling itself from within the same function: in Matlab, functions and classes inside packages are never in scope by default, not even within code inside those same packages. So they must be package-qualified or `import`ed.


Remi-Gau:
This looks good to me and I used the occasion to add a test for tsvread. 

(unrelated: @apjanke why could I directly push to your branch? I thought I would have to make a PR on it for my changes to be added? Git is still a mystery to me. If you want to send a proper PR let me know. Sorry about that.)

apjanke:
> why could I directly push to your branch? 

This is a new GitHub feature introduced about a year ago, where you can optionally let maintainers of the repo you are PRing to directly modify the PR of your branch, instead of going through the sub-PR process. It's controlled by this "Allow edits from maintainers" checkbox on the PR page:

Screen Shot 2019-12-23 at 2 53 54 PM

I'm fine with you adding commits here; no need to do a separate PR.

I've rebased this PR on `master` to resolve the merge conflicts.

Remi-Gau:
> This is a new GitHub feature introduced about a year ago, where you can optionally let maintainers of the repo you are PRing to directly modify the PR of your branch, instead of going through the sub-PR process. It's controlled by this "Allow edits from maintainers" checkbox on the PR page:

I had seen this tick box but I had not realized what it did in practice so I would usually still to a PR. Here I forgot about it and did git push and it worked and I went... That's not supposed to happen.

Thanks for teaching me.

Issue #65, opened 2020-10-25
Querying for modalities does not filter for sessions or subjects


Remi-Gau:
Example from the 7T-trt dataset from bids-example

```
$ tree -d sub-0[12]
sub-01
├── ses-1
│   ├── anat
│   ├── fmap
│   └── func
└── ses-2
    ├── fmap
    └── func
sub-02
├── ses-1
│   ├── anat
│   ├── fmap
│   └── func
└── ses-2
    ├── fmap
    └── func
```

### Actual behavior
```
BIDS = bids.layout(dataset_path);
bids.query(BIDS,  'modalities',  'sub',  '01',  'ses',  '2')

ans =

  1×3 cell array

    'anat'    'func'    'fmap'
```

### Expected behavior
```
BIDS = bids.layout(dataset_path);
bids.query(BIDS,  'modalities',  'sub',  '01',  'ses',  '2')

ans =

  1×2 cell array

   'func'    'fmap'
```

### Notes
This also applies to `spm_bids`

gllmflndn:
Confirmed. It is similar to #12: `subjects`, `sessions` and `modalities` were dealt with differently than the other keywords - they are qualitatively different but it seems that the code path can be the same. `modalities` is the [last one](https://github.com/bids-standard/bids-matlab/blob/master/%2Bbids/query.m#L31) treated separately.
I will open a pull request with a potential fix.

Remi-Gau:
Fixed by PR #71

PR #71, opened 2020-10-27
[FIX] Filter subjects and sessions when querying modalities (issue #65)

gllmflndn:
It might be time to remove the remaining `switch`/`case` as there are no special cases any more.

### DIFF
--- a/+bids/query.m
+++ b/+bids/query.m
@@ -143,7 +143,7 @@ switch query
                 result = unique(result);
                 result = regexprep(result,'^[a-zA-Z0-9]+-','');
                 result(cellfun('isempty',result)) = [];
-            case 'data'
+            case {'modalities','data'}
                 result = result';
             case 'metadata'
                 if numel(result) == 1
--- a/+bids/query.m
+++ b/+bids/query.m
@@ -28,13 +28,7 @@ BIDS = bids.layout(BIDS);
 opts = parse_query(varargin);
 
 switch query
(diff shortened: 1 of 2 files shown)


Remi-Gau:
And here a sentence you are going to hear me say often: "Can I add a test for that?" :wink:

Hoping that bit by bit we can increase our coverage.

gllmflndn:
> And here a sentence you are going to hear me say often: "Can I add a test for that?" 😉

I thought you volunteered for that? :laughing: Otherwise I nominate @ChristophePhillips

Remi-Gau:
> > And here a sentence you are going to hear me say often: "Can I add a test for that?" wink
> 
> I thought you volunteered for that? laughing Otherwise I nominate @ChristophePhillips

I wlll gladly do it but if @ChristophePhillips wants to have fun at it: be my guest.

Remi-Gau (approved):

Remi-Gau:
> It might be time to remove the remaining `switch`/`case` as there are no special cases any more.

This looks good to me and yes that external switch seems a bit of an overkill though I think we need to have the error handling at the end.

gllmflndn:
Thanks, I will leave the switch for now as a reminder of the former distinction between different queries (this might require further thoughts).
Also, the tests with `isequal` enforce a particular order, e.g. `{'anat','fmap','func'}`: we should probably relax this at some point.

PR #80, opened 2020-11-01
add code of conduct from bids-specs

Remi-Gau:
This is not meant to be a final version in any way but having something is better than nothing. We can try to improve and go beyond the COC of bids-specs with time.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of
+experience, nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+- Using welcoming and inclusive language
+- Being respectful of differing viewpoints and experiences
+- Gracefully accepting constructive criticism
+- Focusing on what is best for the community
+- Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+- The use of sexualized language or imagery and unwelcome sexual attention or
+  advances
+- Trolling, insulting/derogatory comments, and personal or political attacks
+- Public or private harassment
+- Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+- Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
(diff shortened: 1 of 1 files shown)


gllmflndn:
Concerning a Code of Conduct, I leave BIDS maintainers @Remi-Gau @robertoostenveld @sappelhoff decide which one is most appropriate for projects hosted on @bids-standard.

robertoostenveld:
I believe there is a general one that applies to all repositories, unless explicitly overruled by a specific one. Unless there are good reasons otherwise, I suggest to stick to that one. 

See https://github.com/bids-standard/.github/blob/master/CODE_OF_CONDUCT.md. I must admit that I don't know how that actually works, i.e. how the one from `.github` is made to apply here.

sappelhoff:
> See https://github.com/bids-standard/.github/blob/master/CODE_OF_CONDUCT.md. I must admit that I don't know how that actually works, i.e. how the one from .github is made to apply here.

when opening a new issue or PR, you will see a link to the bids-standard "default" COC on the lower right, see this screenshot:

grafik

---

It's not very visible unfortunately ... which is why I would be okay with adding it explicitly in this repo (as is also done at the BIDS-specification repo). See https://github.com/bids-standard/bids-specification/issues/364 for more information.

apjanke:
> I believe there is a general one that applies to all repositories, unless explicitly overruled by a specific one.
> [...]
> See https://github.com/bids-standard/.github/blob/master/CODE_OF_CONDUCT.md. I must admit that I don't know how that actually works, i.e. how the one from .github is made to apply here.

I do not think that is the case. This thing you have linked to is a repo named ".github" - that's not a standard GitHub feature as far as I know, and I do not think it has any effect. The ".github" functionality is a per-repo thing, where you create a directory named `.github` inside each repo. I do not think creating a repo named `.github` has any effect, and I don't think there's any "inheritance" from it to the individual organizations inside your repo.

For example, when I view the `bids-matlab` repo, or create a new PR against `bids-matlab`, I do not see any banners or links for a code of conduct there.

This is what I see when looking at the bids-matlab repo:

image

And this is what I see when I create a new PR for the bids-matlab repo:

image

I don't see any CoC references anywhere in them.

I don't see the "Helpful Resources" section that @sappelhoff saw in their PR. (Maybe you could include the URL bar to make it clearer what you're making a PR against there?)

I think you need to add the CoC to each individual repo to have it take effect.

Remi-Gau:
The link is there: below the comment box at the bottom.

```
Remember, contributions to this repository should follow its code of conduct. 
```

:arrow_down:

apjanke:
Ohhh! My eyes slid *right* over that, even after several re-scans of the page; that's the section for "fine print that I've trained myself to completely ignore".

I kind of feel like a doofus now, but also think this maybe means it should be featured somewhere more prominent, if it could be.

Remi-Gau:
The problem is not with you definitely. 

So that speaks in favor of having a COC in the repo I think. :-)

Remi-Gau:
No veto on this until now so I am merging.

At least this gives a basis to improve on.

PR #86, opened 2020-11-11
Add .editorconfig?

apjanke:
What do you think about adding an `.editorconfig` file to make it easier for contributors to stick to the repo's style guidelines?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.m]
+indent_style = space
+#indent_size = 2
+
+[*.yml]
+indent_style = space
+indent_size = 2


Remi-Gau:
Good idea: need to check how it plays with the `miss_hit` linter we have put in place.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.m]
+indent_style = space
+indent_size = 4
```
Remi-Gau, line 9:
```suggestion
indent_size = 2
```

Just to make that consistent with the `miss_hit` configuration.

https://github.com/bids-standard/bids-matlab/blob/master/miss_hit.cfg

Remi-Gau (changes requested):
tiny detail to make it play well with our linter

.editorconfig
apjanke: Sounds fine to me, but all the existing `*.m` files in the repo seem to use 4-space indents, not 2-space?

Remi-Gau: Yes. you are correct. 

The PR that set up the linter did not apply the automatic linting of miss_hit yet. Once again this was a conscious choice to make that process happen bit by bit because otherwise it meant changing all the lines in all `.m` functions and potentially leading to a plethora of merge conflicts with other PRs. 

https://github.com/bids-standard/bids-matlab/pull/58#issuecomment-723417158

The idea is to run `mh_style function_to_lint --fix` on the functions changed in a PR as a last commit before merge to make review easier and try to minimize "meaning less" conflict merges.

Does that make sense ?

apjanke: I think maybe `miss_hit.cfg` should be changed to `tab_width: 2`. There's 3,000 style flags from `mh_style`:
```
mh_style | tail -1
MISS_HIT Style Summary: 26 file(s) analysed, 3336 style issue(s)
```
and about 500 of them look like complaints about indentation length. If I change it to `tab_width: 2` it drops to 2,851 style issues.

apjanke: Oh, yeah, that makes sense. I'll make the change here. And maybe we should just leave `indent_size = 2` commented out here until it's actually being enforced?

Remi-Gau: Yes! Let's do that!! :smile:

apjanke: Revised the PR to do so.

Remi-Gau (approved):
Awesome ! Thanks @apjanke for this!

Remi-Gau:
@all-contributors please add @apjanke for infra

apjanke:
You're welcome!

PR #87, opened 2020-11-11
miss_hit.cfg: define project_root

apjanke:
The [MISS_HIT Configuration Guide](https://florianschanda.github.io/miss_hit/configuration.html) recommends putting a `project_root` in your top-level `miss_hit.cfg` file. This will make it play nicely if bids-matlab is included in some other project using MISS_HIT, and maybe optimize the `mh_*` programs' lookup of config files.

### DIFF
--- a/miss_hit.cfg
+++ b/miss_hit.cfg
@@ -1,5 +1,7 @@
 # style guide (https://florianschanda.github.io/miss_hit/style_checker.html)
 
+project_root
+
 # rules that we do not apply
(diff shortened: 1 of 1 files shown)


Remi-Gau (approved):
Good catch. 

I have to admit I need to go for a deeper dive on the miss_hit linter. :smile:

Remi-Gau:
This is infrastructure related so I will merge right away.

PR #94, opened 2020-11-11
Skip datasets when labelled as non-conformant

gllmflndn:
Due to new derivatives dataset added in bids-examples:
https://github.com/bids-standard/bids-examples/tree/master/ds000001-fmriprep


Remi-Gau (approved):
I did not even know about `.SKIP_VALIDATION`.

Learning things every day.

gllmflndn:
Me neither... Not sure it's a real thing but at least the tests will pass again for now.

PR #95, opened 2020-11-12
apply MISS_HIT linter across the board

Remi-Gau:
This is a first pass at fixing most of the issues on files that are not involved in another PR (pretty much everything but the report function).

### For reviewers

The only thing that is not purely aesthetic is a small refactoring of `layout.m` where the entities for each modality have been placed in a cell and used in each "section" related to that modality.

It will be easier to see on the second commit of that PR.

https://github.com/bids-standard/bids-matlab/commit/4dfa649741e98a253444541a64eab21dde5d2046#diff-f8205f26841517365010841c7994796349b7d7dc6b5708b8a0c3d6962e04b34dL212


Remi-Gau:
At the moment there are only 5 linting issues left in `layout.m` and most of the other issues in `report.m` should be addressed in  #45 and can be autofixed.

apjanke:
LGTM, based on skimming the diff.

This is just a personal aesthetic judgment, but the new format is much easier on the eyes for me.

I'm impressed with what MISS_HIT has been able to do here. Probably time to add that to my own personal toolbox.

PR #99, opened 2020-11-28
[FIX] add "dir" to the list of func entities

Remi-Gau:
`dir` was not listed as one of the entities for the `func` modalities.

This led to some bug on our side: https://github.com/cpp-lln-lab/CPP_SPM/issues/194

The `func` name template is this one:

```
sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_<contrast_label>.nii[.gz]
```

The current entities we support are:
```
    func_entities = {'sub', ...
                     'ses', ...
                     'task', ...
                     'acq', ...
                     'rec', ...
                     'fa', ...
                     'echo', ...
                     'inv', ...
                     'run', ...
                     'recording', ...
                     'meta'};
```

I see quite a few other mismatch in there. @gllmflndn am I missing something or shouldn't we also:

- add `ce`
- remove `fa`, `inv`, `recording`


Remi-Gau:
Also we only support bold as contrast label for now and `cbv` and `phase` are also in the specs. 

This seems more of an edge case at the moment and might require more coding that could be rendered moot when we have a schema to help us. So I would suggest that we open an issue for this point (to remind ourselves and let others know we know) but do not move on to fix it right away.

Remi-Gau:
Unless I am mistaken there is no dataset in bids-example with a `dir` entity so there is no convenient way to add a test for this one.

So I will merge for now and make a note to add a test for this... Later.

Issue #68, opened 2020-10-26
Move tests from Travis to GitHub Actions to test on all OSses


sappelhoff:
GitHub Actions allows for easy testing on Windows, MacOS, and Linux, which would improve the testing pipeline to also cover "special" bugs that only appear on some OSses.

Remi-Gau:
Hum... Even with matlab and octave?

I know that Travis actually now supports Matlab in their CI so we could have both there. Not sure about github actions.

sappelhoff:
> Hum... Even with matlab and octave?

Octave yes (not sure about Windows to be honest) - but Matlab probably not. I didnt know that Travis supports Matlab in their CI, that's awesome ... and a bit puzzling to me how they deal with the licensing issues :)

Remi-Gau:
Well I don't know much more than that: https://twitter.com/RemiGau/status/1310681042714537991
And I have not tested it yet

sappelhoff:
I now found this: https://github.com/joergbrech/moxunit-action

but haven't tested it yet.

> This action performs unit tests for GNU Octave and **Matlab** using the unit testing framework MOxUnit. 

Looks like we need some time and energy to test a couple of alternatives :-)

Remi-Gau:
Ha yeah we started thinking about using MOxUnit in #44 .

No decision has been made about this yet. But it definitely is an option that can also get you your code coverage in one fell swoop.

gllmflndn:
In general, I would also prefer to use GitHub Actions to anything else but if we want to test with MATLAB and not just Octave, we might have to (temporarily) still use another platform. This [page](https://www.mathworks.com/solutions/continuous-integration.html) only mentions CircleCI and Travis, while [this one](https://www.mathworks.com/help/matlab/matlab_prog/continuous-integration-with-matlab-on-ci-platforms.html) does not mention [Travis](https://docs.travis-ci.com/user/languages/matlab/).

Remi-Gau:
@sappelhoff 

any pointer on trying to adapt our github action workflows to run our tests also on MacOS and Windows (for octave)?

I don't have the bandwidth to look on setting up CI tests with matlab at the moment.

sappelhoff:
Yes, I think setting up for at least MacOS will be relatively straight forward: You will just have to make another step like this one:

https://github.com/bids-standard/bids-matlab/blob/6c1dac65880df28b69ec131983cd8bbbfce1c258/.github/workflows/run_tests.yml#L22-L26

[...]

PR #100, opened 2020-12-28
[INFRA] run tests with github action

Remi-Gau:
- drop travis for gtihub actions


Remi-Gau:
fix #68

Remi-Gau:
Hum... Unclear as to why this action fails here but seemed to work fine on a PR on my own repo...

https://github.com/Remi-Gau/bids-matlab/pull/5

gllmflndn:
Thanks for doing this @Remi-Gau. I temporarily forced the tests to use [ubuntu-20.04](https://github.blog/changelog/2020-10-29-github-actions-ubuntu-latest-workflows-will-use-ubuntu-20-04/) and they ran through. We can revert to `ubuntu-latest` when the switch is over.
It's probably just me but I would configure miss_hit to be informative (i.e. report full list of detected issues) but not fail as a test while the library is under heavy development (i.e. implements a should rather than a must).

Remi-Gau:
> Thanks for doing this @Remi-Gau. I temporarily forced the tests to use [ubuntu-20.04](https://github.blog/changelog/2020-10-29-github-actions-ubuntu-latest-workflows-will-use-ubuntu-20-04/) and they ran through. We can revert to `ubuntu-latest` when the switch is over.

Cool thanks!!

> It's probably just me but I would configure miss_hit to be informative (i.e. report full list of detected issues) but not fail as a test while the library is under heavy development (i.e. implements a should rather than a must).

yeah it's just you!!! :wink: (that's my way of saying "happy new year!!!"! Have you missed me?)

I was thinking of a "smart" way of doing this (because I don't think that miss_hit has "warning only" mode at the moment): to align with the idea of developing stuff on a `dev` branch and keep the `master` branch, we could from now on work on the `dev` and configure the github action to run miss_hit only on `master` or PRs to master (I hope that github action allows to only run stuff on PR on a certain base branch).

Would that make sense?

Remi-Gau:
https://stackoverflow.com/questions/63817052/github-actions-run-step-only-for-certain-pull-request-base-branches

Need to double check but something like this could do the trick:

```yaml
on:
  push:
    branches: [master, main]
  pull_request:
    branches: [master, main]
```

@gllmflndn  Would you be OK with that?

Remi-Gau:
Unless I hear otherwise I will merge this in 2 days: this is mostly infrastructure related but it should make our lives easier moving forward.

sappelhoff (approved):

PR #109, opened 2021-01-13
refactoring and linting

Remi-Gau:
my favorite things...

- [x] (aggressive) refactoring of the layout function
- [x] linting
- [x] add test for `file_utils`
- [x] refactoring of `file_utils` to make it easier to read


Remi-Gau:
Note to self: Why isn't Travis running??

Remi-Gau:
so more work can still be done but I will stop here for now.

I suspect that some of this could be useful when we start implementing the schema.

Remi-Gau:
Unless I hear otherwise I will merge this in 2 days: this is mostly infrastructure related but it should make our lives easier moving forward.

PR #123, opened 2021-02-04
[ENH] query returns events file metadata and content

Remi-Gau:
test to confirm issue raised in #144


Remi-Gau:
will not implement unless someone really asks for it

PR #124, opened 2021-02-04
[ENH] implement bids-schema - part 1

### LINKED ISSUE #103: Add support for  `cbv` and `phase` sufixes for `func` files
Also we only support `bold` as contrast label for now and `cbv` and `phase` are also in the specs. 

This seems more of an edge case at the moment and might require more coding that could be rendered moot when we have a schema to help us. So I would suggest that we open an issue for this point (to remind ourselves and let others know we know) but do not move on to fix it right away.

_Originally posted by @Remi-Gau in https://github.com/bids-standard/bids-matlab/issues/99#issuecomment-735287218_

### LINKED ISSUE #104: Missing or extra entities for `func`
This should be resolved when we implement the schema but I am opening an issue for the record.

- add `ce` entity
- remove `fa`, `inv`, `recording` entities

Remi-Gau:
Use the bids-schema from the specification repo to parse datasets.

- [x] create function to load schema
- [x] implement file parsing according to datatypes and modalities of the schema
- [x] add functions to handle (for a given datatype):
  - [x] entities
  - [x] suffixes
  - [x] extensions
- [x] implement schema for parsing
  - [x] anat
  - [x] behavioral
  - [x] ASL
  - [x] dwi
  - [x]  func (fixes #104 and fixes #103)
  - [x] eeg
  - [x] meg
  - [x] ieeg
  - [x] fmap
- [x] update field and variable names to fit with the bids terminology (ex: "suffix" instead of "type")
- [x] support schema less parsing and minimal support of derivatives
- [x] allow query to return file dependency information

# Issues

- Some suffixes are deprecated in BIDS 1.5 and not included in the schema (e.g `FLASH`), and therefore will be ignored by `layout` when parsing them.

- Different types of "suffixes" allow for different set of entities for the same datatype (see below T1w do have a part entity and T1map does not, even though they are both `anat`). Relying on the schema when parsing leads to create concatenating structures that have different set of fields.
  - FIXED but creates structure that combine entities from different suffixes that are not ordered.
  
 - The schema has to be loaded several times when parsing a dataset: this can seriously slow things down (especially on Octave and especially when running the tests where all `bids-examples` have to be parsed.
   - FIXED though things are still very slow if a lot of datasets have to be parsed.

```
  {
    "suffixes": [
      "T1w",
    ],
    "entities": {
      "subject": "required",
      "session": "optional",
      "run": "optional",
      "acquisition": "optional",
      "ceagent": "optional",
      "reconstruction": "optional",
      "part": "optional"
    }
  },
  {
    "suffixes": [
      "T1map",
    ],
    "entities": {
      "subject": "required",
      "session": "optional",
      "run": "optional",
      "acquisition": "optional",
      "ceagent": "optional",
      "reconstruction": "optional"
    }
  },

### CODE REVIEW

**Comment by Remi-Gau** [comment]
CI fails for several errors of the same type.

Nothing similar locally with Octave 4.2, 5.1 or Matlab

```
failure: cell cannot be indexed with .
  return_datatype_suffixes:5 (/github/workspace/+bids/+internal/return_datatype_suffixes.m)
  test_return_datatype_suffixes>test_return_datatype_suffixes_basic:13 (/github/workspace/tests/test_return_datatype_suffixes.m)
```  

```
function suffixes = return_datatype_suffixes(datatype)

  suffixes = '_(';

  for iExt = 1:numel(datatype.suffixes)   % < --------------------------------- problem is here (???)
    suffixes = [suffixes,  datatype.suffixes{iExt}, '|']; %#ok<AGROW>
  end

  % Replace final "|" by a "){1}"
  suffixes(end:end + 3) = '){1}';

end
```

Remi-Gau:
[OK I give up](https://www.urbandictionary.com/define.php?term=FTS)...

... for now!

Remi-Gau:
Fixed problem with CI.

Had to convert cells to structures in several places just for CI.
Not sure of the source of the problem for the difference in behavior though, though I suspect the way JSONs are read in CI gives the JSON schema a different structure.

Remi-Gau:
Will merge this now before this PR grows even more out of control.
But will first make a list of the features that should be added after this.

Remi-Gau:
- [ ] allow query to return info about the extension (is table? is zipped?)

Remi-Gau:
- [ ] schema-less layout also indexes also json files

Remi-Gau:
[ ] generalize usage of `intended_for` field where applicable and `dependency` sub-structure

Remi-Gau:
- [ ] query should return events file metadata and content

PR #151, opened 2021-02-16
[ENH] improve management of "intended_for"

Remi-Gau:
- [x] anything to do with files with a potential `intendedFor` metadata field is done after the index of the dataset
- [x] the `intended_for` field of file X contains the fullpath of all the existing target files that X is actually intended for
- [x] each target file of the `intendedFor` list of file X get an `informed_by` field with the fullpath for X

Does not include a way to query the `intended_for` or `informed_by` field.

---

For reviewers:

This is more on the implementation side, but in terms of "output" I am wondering what would the most useful for users. Ideally, this approach would then be extended to `events` and other "associated files".

- should `informed_by` (pybids uses this I think) replace the `dependencies` field? Which one is more "intuitive"?

See example below

```matlab
BIDS = bids.layout(fullfile(pth_bids_example, 'asl004'));

>> BIDS.subjects.fmap

ans = 

  struct with fields:

        filename: 'sub-Sub1_dir-pa_m0scan.nii.gz'
             ext: '.nii.gz'
          suffix: 'm0scan'
        entities: [1×1 struct]
            meta: [1×1 struct]
    intended_for: {'/home/remi/github/BIDS-matlab/tests/bids-examples/asl004/sub-Sub1/perf/sub-Sub1_m0scan.nii.gz'}


>> BIDS.subjects.perf(1)

ans = 

  struct with fields:

        filename: 'sub-Sub1_asl.nii.gz'
             ext: '.nii.gz'
          suffix: 'asl'
        entities: [1×1 struct]
            meta: [1×1 struct]
    dependencies: [1×1 struct]
     informed_by: [1×1 struct]
    intended_for: []

>> BIDS.subjects.perf(1).informed_by.perf

ans =

    '/home/remi/github/BIDS-matlab/tests/bids-examples/asl004/sub-Sub1/perf/sub-Sub1_m0scan.nii.gz'

>> BIDS.subjects.perf(4)

ans = 

  struct with fields:

        filename: 'sub-Sub1_m0scan.nii.gz'
             ext: '.nii.gz'
          suffix: 'm0scan'
        entities: [1×1 struct]
            meta: [1×1 struct]
    dependencies: ''
     informed_by: [1×1 struct]
    intended_for: {'/home/remi/github/BIDS-matlab/tests/bids-examples/asl004/sub-Sub1/perf/sub-Sub1_asl.nii.gz'}


>> BIDS.subjects.perf(4).informed_by.fmap

ans =

    '/home/remi/github/BIDS-matlab/tests/bids-examples/asl004/sub-Sub1/fmap/sub-Sub1_dir-pa_m0scan.nii.gz'
```

### DIFF
--- a/+bids/+internal/get_metadata.m
+++ b/+bids/+internal/get_metadata.m
@@ -1,8 +1,17 @@
 function meta = get_metadata(filename, pattern)
+  %
   % Read a BIDS's file metadata according to the inheritance principle
-  % FORMAT meta = bids.internal.get_metadata(filename, pattern)
-  % filename    - name of file following BIDS standard
-  % pattern     - regular expression matching metadata file
+  %
+  % USAGE::
+  %
+  %    meta = bids.internal.get_metadata(filename, pattern)
+  %
+  % :param filename: name of file following BIDS standard
+  % :type  filename: string
+  % :param pattern:  regular expression matching metadata file
+  % :type  pattern:  string
+  %
+  %
   % meta        - metadata structure
   % __________________________________________________________________________
 
--- a/+bids/+internal/get_metadata.m
+++ b/+bids/+internal/get_metadata.m
@@ -4,11 +4,12 @@ function meta = get_metadata(filename, pattern)
   %
   % USAGE::
   %
-  %    meta = bids.internal.get_metadata(filename, pattern)
+  %    meta = bids.internal.get_metadata(filename, pattern = '^.*%s\\.json$')
   %
-  % :param filename: name of file following BIDS standard
+  % :param filename: fullpath name of file following BIDS standard
   % :type  filename: string
-  % :param pattern:  regular expression matching metadata file
+  % :param pattern:  Regular expression matching the metadata file (default is ``'^.*%s\\.json$'``)
+  %                  If provided, it must at least be ``'%s'``.
   % :type  pattern:  string
   %
   %
--- a/+bids/layout.m
+++ b/+bids/layout.m
@@ -243,6 +243,7 @@ function subject = parse_dwi(subject, schema)
         % bval & bvec file
         % ------------------------------------------------------------------
         % TODO: they can also be stored at higher levels (inheritance principle)
+        % TODO: refactor and deal with all this after file indexing
         bvalfile = bids.internal.get_metadata(fullpath_filename, '^.*%s\\.bval$');
         if isfield(bvalfile, 'filename')
           subject.dwi(end).dependencies.bval = bids.util.tsvread(bvalfile.filename);
@@ -476,6 +477,12 @@ end
 
 function BIDS = manage_intended_for(BIDS)
 
+  % Loops through all the files with potential ``intentedFor`` metadata
+  % and creates an ``intended_for`` field with the fullpath list of all the target files
+  % it is intended for that exist.
+  %
+  % Also update the structure of each target file with an ``informed_by`` field
+
   suffix_with_intended_for = { ...
                               'phasediff'; ...
                               'phase1'; ...
@@ -503,9 +510,14 @@ function BIDS = manage_intended_for(BIDS)
           intended_for;
 
       % update "dependency" field of target file
+      informed_by = [];
       for iTargetFile = 1:size(intended_for, 1)
         info_tgt = bids.internal.return_file_info(BIDS, intended_for{iTargetFile});
 
+        % TODO: this should probably check that the informed_by field does not
+        % already exist in the structure
+        % TODO: This assumes that a file will only be informed by a single file from
+        % this moodality
         informed_by.(info_src.modality) = file_list{iFile};
         BIDS.subjects(info_tgt.sub_idx).(info_tgt.modality)(info_tgt.file_idx).informed_by = ...
             informed_by;
--- a/+bids/layout.m
+++ b/+bids/layout.m
@@ -468,17 +468,21 @@ end
 function structure = manage_tsv(structure, pth, filename)
 
   ext = bids.internal.file_utils(filename, 'ext');
-  p = bids.internal.file_utils('FPList', pth,  ['^' strrep(filename, ['.' ext], ['\.' ext]) '$']);
+  tsv_file = bids.internal.file_utils('FPList', ...
+                                      pth,  ...
+                                      ['^' strrep(filename, ['.' ext], ['\.' ext]) '$']);
 
-  if isempty(p)
+  if isempty(tsv_file)
     warning('Missing: %s', fullfile(pth, filename));
 
   else
-    structure.content = bids.util.tsvread(p);
+    structure.content = bids.util.tsvread(tsv_file);
 
-    p = bids.internal.file_utils('FPList', pth,  ['^' strrep(filename, ['.' ext], '\.json') '$']);
-    if ~isempty(p)
-      structure.meta = bids.util.jsondecode(p);
+    tsv_file = bids.internal.file_utils('FPList', ...
+                                        pth,  ...
+                                        ['^' strrep(filename, ['.' ext], '\.json') '$']);
+    if ~isempty(tsv_file)
+      structure.meta = bids.util.jsondecode(tsv_file);
     end
 
   end
--- a/+bids/layout.m
+++ b/+bids/layout.m
@@ -512,6 +512,7 @@ function BIDS = manage_intended_for(BIDS)
       % update "dependency" field of target file
       informed_by = [];
       for iTargetFile = 1:size(intended_for, 1)
+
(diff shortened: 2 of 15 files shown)


+bids/+internal/get_metadata.m
```diff
@@ -1,8 +1,17 @@
...
-  % filename    - name of file following BIDS standard
-  % pattern     - regular expression matching metadata file
+  %
+  % USAGE::
+  %
+  %    meta = bids.internal.get_metadata(filename, pattern)
+  %
+  % :param filename: name of file following BIDS standard
+  % :type  filename: string
+  % :param pattern:  regular expression matching metadata file
```
nbeliy, line 11: The docstring is a little bit confused. Reading from code, filemane must be datafile, and pattern must contain `%s`. This must appear in docstring.

+bids/layout.m
```diff
@@ -456,66 +453,120 @@ function tolerant_message(use_schema, msg)
...
 
   else
-    structure.content = bids.util.tsvread(p);
+    structure.content = bids.util.tsvread(tsv_file);
+
+    tsv_file = bids.internal.file_utils('FPList', ...
+                                        pth,  ...
+                                        ['^' strrep(filename, ['.' ext], '\.json') '$']);
+    if ~isempty(tsv_file)
+      structure.meta = bids.util.jsondecode(tsv_file);
```
nbeliy, line 471: If both tsv and json are readed into memory, it would be worth to check their consistency -- ie if all columns in tsv are defined in json.

+bids/layout.m
```diff
@@ -456,66 +453,120 @@ function tolerant_message(use_schema, msg)
...
+                                      pth,  ...
+                                      ['^' strrep(filename, ['.' ext], ['\.' ext]) '$']);
 
-  if isempty(p)
+  if isempty(tsv_file)
     warning('Missing: %s', fullfile(pth, filename));
 
   else
-    structure.content = bids.util.tsvread(p);
+    structure.content = bids.util.tsvread(tsv_file);
```
nbeliy, line 465: Don't see much point to store content of tsv file in structure -- it can be large and their treatment is very much contextual.

+bids/layout.m
```diff
@@ -456,66 +453,120 @@ function tolerant_message(use_schema, msg)
...
 
-      else
-        % also check that this NIfTI aims to the same m0scan
-        [~, path2check, ext2check] = fileparts(path_intended_for{iPath});
-        filename_found = max(arrayfun(@(x) strcmp(x.filename, ...
-                                                  [path2check ext2check]), ...
-                                      subject.perf));
-        if ~filename_found
-          warning(['Did not find NIfTI for which is intended: ' structure.filename]);
+    file_path = strrep(subject.path, subject.session, '');
```
nbeliy, line 559: Unsafe: If `subject.path` contains two or more occurrences of `subject.session`, they will be both replaced, making path unusable. Only last occurrence must be replaced.

Also in bids intended for paths are stored in unix way, with `/` separator. The code may be run on windows system with `/` separators. I'm not sure that it will not cause problems during path generation.

+bids/query.m
```diff
@@ -257,7 +257,7 @@
 
             case 'dependencies'
               if isfield(d(k), 'dependencies')
-                result{end + 1} = d(k).dependencies;
+                result{end + 1, 1} = d(k).dependencies;
```
nbeliy, line 260: In general query returns a array of strings, except for dependencies, in which case it will return an array of structure.

+bids/layout.m
Remi-Gau: Yeah I was not super enthusiastic about this and it can potentially slow down the layout indexing if you have a large dataset.

Will remove it.

+bids/query.m
Remi-Gau: Yeah metadata is also handled that way.

If you ask info about a single file then you get a structure.

If you ask for metada or dependencies about several files, you can get structures that have different shapes so you can't concatenate them in something prettier and easier to handle. :sob:

+bids/layout.m
Remi-Gau: Yeah not sure about especially if we drop reading the files during indexing (see below).

+bids/layout.m
Remi-Gau: > Also in bids intended for paths are stored in unix way, with `/` separator. The code may be run on windows system with `/` separators. I'm not sure that it will not cause problems during path generation.

Good catch. Will update that.

Remi-Gau: > Unsafe: If `subject.path` contains two or more occurrences of `subject.session`, they will be both replaced, making path unusable. Only last occurrence must be replaced.

Ha yes! Thanks!

+bids/+internal/get_metadata.m
Remi-Gau: Oh nice. I am slowly trying to improve the "help" sections and reformating them to use sphynx in the end to generate the doc. But I am clearly lagging behind in keeping that up to date.

+bids/layout.m
nbeliy: It is question of design. My opinion (purely subjectif):
For layout metadata is needed only for parcing IntendedFor fields
For query returning full structure and just path to json file are equivalent, retrieving full stucture is trivial using `bids.util.jsondecode`, hence no need to return full structure.
But, what can be useful is to query on individual fields of meta, i.e. "show me all files with repetition time == 0.03". Then contains of meta will become important for query. But it is difficult to implement (espetualy if queried values are numbers and not strings).

The question if load in memory all meta, or will read given file each time we look in it -- is open. One economise time, other memory. For me it is better to read each time -- layout and query will be called once, and before processing images which will dominate calculation time. The json files are small, so user will see slow-down only in huge datasets (but there any solution will be slow).

For indexing, I can't comment beacause of lack of matlab knowelege

+bids/layout.m
Remi-Gau: OK I will start by removing the reading of the TSV content upon indexing for most modalities.

I am keeping some of it for `dwi` and `perf` **for now**. 

Most likely all tsv reading will be moved to the `query` side of things: will do that in several steps.

The good news is that tests run MUCH faster already.

+bids/layout.m
Remi-Gau: > It is question of design. My opinion (purely subjectif):
:wink: 

> For layout metadata is needed only for parcing IntendedFor fields
> For query returning full structure and just path to json file are equivalent, retrieving full stucture is trivial using `bids.util.jsondecode`, hence no need to return full structure.

Better to use `get_metadata` that should take care of the inheritance principle but I see what you mean.

> But, what can be useful is to query on individual fields of meta, i.e. "show me all files with repetition time == 0.03". Then contains of meta will become important for query. But it is difficult to implement (espetualy if queried values are numbers and not strings).

There is some code "ruins" in `bids.query` about this: so I will leave that there for though I agree that doing this on query could be painful.

> The question if load in memory all meta, or will read given file each time we look in it -- is open. One economise time, other memory. For me it is better to read each time -- layout and query will be called once, and before processing images which will dominate calculation time. The json files are small, so user will see slow-down only in huge datasets (but there any solution will be slow).

I agree that the bottle neck in our pipeline is definitely not json metadata reading. And when moving to big data set, we might have to go a completely different way anyway.

nbeliy (approved):
Not fun of making each dependency in its own field, but I can see the logic behind.

Remi-Gau:
> Not fun of making each dependency in its own field, but I can see the logic behind.

yeah I agree that this is a bit cumbersome, if we find a more elegant solution down the line let's see how to improve.

nbeliy:
Sorry to reopening the request. There two issues that I encountered during the tests:

1. Do not take into account qMRI fieldmaps (TB1EPI etc...)
2. The json list in IntendedFor is transformed into cell and not structure, hence it avoid  elseif on line 527
```
ans =

  6x1 cell array

    {'anat/sub-HC1278_echo-*_acq-T1w_part-phase_MPM.nii.gz'  }
    {'anat/sub-HC1278_echo-*_acq-T1w_part-mag_MPM.nii.gz'    }
    {'anat/sub-HC1278_echo-*_acq-MToff_part-mag_MPM.nii.gz'  }
    {'anat/sub-HC1278_echo-*_acq-MToff_part-phase_MPM.nii.gz'}
    {'anat/sub-HC1278_echo-*_acq-MTon_part-mag_MPM.nii.gz'   }
    {'anat/sub-HC1278_echo-*_acq-MTon_part-phase_MPM.nii.gz' }
```

Proposition of solution (not sure if it will broke something else):
1a) Search IntendedFor in each of json files
1b) Search IntendedFor in each file in modality fmap, not sure how it will work for coodsystem and m0

Remi-Gau:
@all-contributors please add @nbeliy  for review

PR #154, opened 2021-02-16
[ENH] schema-less indexing collects files prefix that can be returned / filtered by query

### LINKED ISSUE #140: Add prefixes to parsed derivative file names
### Add prefixes to parsed derivative file names 

In the SPM, the output files are usually created with a prefix to original file name. For usual names it is an easy way for ordinary files, but it will break the bidsified names, as nothing can go before `sub-`. 
Thus, if user want to daisy-chain processing step maintaining he need rename files after each processing step. 
However, if `bids.layout` can support prefix (SPM allow the setting of prefix, so it can be separated from sub by `_`), then the output of each process can be readded directly by `layout`. Obviously this prefix should be defined only for derivatives.

_Originally posted by @nbeliy in https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-779066961_

Remi-Gau:
Same but on the right base

Fixes #140 

When running with `use_schema = false`, the file listed for a given subject must follow the pattern:

```matlab
'^([a-zA-Z0-9]*)sub-'  subject_label  '.*_([a-zA-Z0-9]+){1}\..*$'
```

`parse_filename` will return an extra `prefix` field

```matlab
  filename = '../sub-16/anat/swuasub-16_run-1_task-rest_bold.nii';
  output = bids.internal.parse_filename(filename);

  expected = struct( ...
                    'filename', 'swuasub-16_run-1_task-rest_bold.nii', ...
                    'suffix', 'bold', ...
                    'prefix', 'swua', ...
                    'ext', '.nii', ...
                    'entities', struct(...
                                       'swuasub', '16', ...     % <--------- this makes a weird entity, but querying for subject does not rely on it
                                       'run', '1', ...
                                       'task', 'rest'));
```

I am a bit worried about the weird shape that can now be taken by the "`sub`" entity. Not sure if that can cause issues.

It is possible to query for `prefix`

```matlab
bids.query(BIDS, 'prefixes');


  data = bids.query(BIDS, 'data', ...
                    'sub', '01', ...
                    'run', '01', ...
                    'prefix', 'swua');
```


Remi-Gau:
Merging this into the `dev` branch.
 
So schema less indexing of a bids dataset should now be able of picking up any prefix to the filename and to filter those files with query.

PR #160, opened 2021-02-18
[FIX] make sure running tsvread then tsvwrite does not change the file content

### LINKED ISSUE #156: bids.util.tsvwrite does not correctly write fields and values from struct to output tsv
Imported a *.tsv file with 6 columns with 11 rows of values (excluding headers) using bids.util.tsvread.

The data is loaded in as a struct with 6 fields (name of headers) each with the size [11,1] double. 

Currently in the code, the [11,1] array gets converted into a cell, and when concatenated into `var` on line 40, this becomes a 2x6 cell. 
The consequence is that only 2x6 are written to the output *.tsv because the cells in the second row of `var` are not unrolled, and hence not looped through when writing to the file.

A quick fix would be to replace the following on line 40: `var = [fn'; horzcat(var{:})];`

Then the correct values are written to the output *.tsv file.

Remi-Gau:
fixes #156 

- [ ] run test on some data from bids example and not an extra tsv file
- [ ] support saving row wise structure? This could be pushed to later.

The classic structure tends to not work where mixing types.
```
  logFile.onset = [2; NaN];
  logFile.trial_type = {'motion_up'; 'static'};
  logFile.duration = [1; 4];
  logFile.speed = [NaN; 4];
  logFile.is_fixation = {'true'; '3'};

  % Leads to trouble
  %   logFile.is_fixation = {'true';3};
  %   logFile.is_fixation = {true;3}; 
```  
For example mixing logical and numeric does not work in a cell: it prints to the file but the test fails afterwards.

But the previous structure format could allow mixing the format of elements that would end up in the same column of the tsv.
```
  logFile(1, 1).onset = 2;
  logFile(1, 1).trial_type = 'motion_up';
  logFile(1, 1).duration = 1;
  logFile(1, 1).speed = [];
  logFile(1, 1).is_fixation = true;

  logFile(2, 1).onset = NaN;
  logFile(2, 1).trial_type = 'static';
  logFile(2, 1).duration = 4;
  logFile(2, 1).speed = 4;
  logFile(2, 1).is_fixation = 3;
```  

So we could support this too.


Remi-Gau:
No rush to make a decision on this. If I don't hear back I will silence the test and we can revisit the issue later. :-)

Issue #161, opened 2021-02-19
Provide easy way to loop over files as structure


nbeliy:
In actual state user have an access either to the flat list of files using query, or to full unfiltered layout. It can be helpfull to provide user a way to loop all datafiles returned by query with access to corresponding structure.

It can be done using already existing indexing:
```
finfo = bids.internal.return_file_info(BIDS, data_list{iFile});
file = BIDS.subject(info.sub_idx).(info.modality)(info.file_idx);
``` 
However `return_file_info` rely on find, and so it is time consuming.

The easiest solution (that I see), is to create a new function `query_struct(BIDS, query)` which will replicate existing `query(BIDS, 'data', query)` but returnsa list of found file structures. To optimize, the `query("data")` will call `query_struct(BIDS, query)` and retrieve the filenames.

Alternatively, an iterator can be implemented:
```
iterator = bids.BEGIN(BIDS, query);
do
  file = iterator.get();
while iterator.next()
```
`BEGIN` creates in memory a list of file infos,
`get` returns corresponding file structure
`next` moves current file info to the next in list

PR #162, opened 2021-02-19
[WIP] Function to copy derivatives

nbeliy:
Added a function`derivate(BIDS, out_path, name, varargin)` which exports the files from current BIDS layout into `out_path/name` folder. In other words exports data into derivatives.

`varargin` are the query options same as in `bids.query`, in fact they are passed directly to it.

To do (**help needed!**):

- [ ] Options -- when created derivatives, there a list of usefull options to fine tune, but adding them will make list of parameters to long. Options that I would like to include are below.
- [ ] Script add a line to `dataset_description` a new structure each time it is run. But on the third run it crashes (see below)
- [ ] Add version of bids-matlab (does it exists?)
- [ ] Optimise the loop on files (see #161)
- [ ] Copy of derivatives. Actually it supports only full paths in `dependencies.*` field.
- [ ] Copy of `participants.tsv` and `scans.tsv` files
- [ ] Tests

### Options

- *unzip* -- if true, then zipped files are unzipped
- *copy\_dependencies* -- if true, the files that current file depend on are copied even if they are not explicitely queried
- *force* -- copy file even if he already exists in output. Mutually exclusve with *copy\_dependencies* to avoid circular copy

### Crashes during `dataset_description` generation:
```
Error using horzcat
Dimensions of matrices being concatenated are not consistent.

Error in bids.derivate (line 68)
    description.GeneratedBy = [description.GeneratedBy pipeline];
```

### DIFF
--- a/+bids/derivate.m
+++ b/+bids/derivate.m
@@ -1,0 +1,138 @@
+function derivatives = derivate(BIDS, out_path, name, varargin)
+  %
+  % Copy selected data from BIDS layout to given derivatives folder,
+  % returning layout of new derivatives folder
+  %
+  % USAGE::
+  %
+  %   derivatives = derivate(BIDS, out_path, ...)
+  %
+  % :param BIDS:     BIDS directory name or BIDS structure (from bids.layout)
+  % :type  BIDS:     (strcuture or string)
+  % :param out_path: path to directory containing the derivatives
+  % :type  out_path: string
+  % :param name:     name of pipeline to use
+  % :type  name:     string
+  %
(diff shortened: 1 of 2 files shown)


Remi-Gau:
Thanks for starting this!!!! 

Will have a look and make suggestions where I can.

One think I can already tell you: we don't have bids-matlab version... YET but the issue has been opened and we should try to have some sort of pre-alpha soon.

Remi-Gau:
Closing as this has been moved to #171

PR #163, opened 2021-02-21
Improvements on dependencies calculations

nbeliy:
#### Metadata treatment
Split `get_metadata` into two functions:

- `metalist = get_meta_list(filename, pattern)` returns list of files satisfying pattern with inheritance principle
- `get_metadata(metafile)` construct meta structure from json files in passed meta-files list (from above function)

Each data file in dataset receives new field `metafile` that contains list of files from `get_meta_list`.

This will facilitate access to metafiles directly. Individual values can be extracted by  user if needed using `get_metadata`.

#### Dependencies

Each file structure contains dependencies sub-structure with guaranteed fields:

- `explicit` -- list of data files containing "IntendedFor" referencing current file.
- `data` -- list of files with same name but different extension. This will combine files that are split in header and data (like in Brainvision), also takes care of bval/bvec files
- `group` -- list of files that have same name except extension and suffix. This will group file that logically need each other, like functional mri and events tabular file. It also takes care of fmap magnitude1/2 and phasediff

Additional custom subfields can be added for given modalities.

This allow to treat `dwi` and `fmap` with generic `parse_using_schema`. `perf` likely also, but I can't certify it.

#### Issues

- [ ] performance. I feel that retrieval of dependencies can be optimised
- [ ] Tests -- was unable to install moxunit
- [ ] Tests on other datasets

### DIFF
--- a/+bids/+internal/append_to_layout.m
+++ b/+bids/+internal/append_to_layout.m
@@ -43,8 +43,6 @@ function subject = append_to_layout(file, subject, modality, schema)
     if ~isempty(p) && strcmp(p.ext, '.json')
       return
     end
-    file
-    p.metafile = bids.internal.get_meta_list(file)
 
   end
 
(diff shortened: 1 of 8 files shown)


Remi-Gau:
@nbeliy Thanks for sending this. I hope to be able to have a look at it towards the end of the week. 
Sorry for the delay

Remi-Gau:
moved to #172

Issue #165, opened 2021-02-22
BIDS directory does not exist on Windows


nbeliy:
This is a weird bug:

When trying loading layout on windows using absolute path:
```
BIDS = bids.layout('D:\Bidsification\Bidsified');
Error using bids.layout (line 61)
BIDS directory does not exist: 'Bidsification\Bidsification\Bidsified'
```

Apparently error comes from `+bids/+internal/file_utils.m`, in particular function `canonicalise_path` removes disk name:
```
bids.internal.file_utils('D:\Bidsification\Bidsified', 'CPath')

ans =

    'Bidsification\Bidsified'
```

The ispc is 1.

Remi-Gau:
This highlights the need to set up some CI for windows given the user base.

gllmflndn:
No finger-pointing here but I think that, in this particular example, it might be better to leave these functions unchanged as a mirror of the battle-tested `spm_select`/`cfg_getfile` files they were copied from, to minimise the maintenance burden.
https://github.com/bids-standard/bids-matlab/commit/30526b9f53c69ad388faf50cc512941f9f3a40c0#diff-801283b7ca73698423622e9903c08949b16555ade75d6268a33a89b3e731bafeR155-L138

Remi-Gau:
> No finger-pointing here but I think that, in this particular example, it might be better to leave these functions unchanged as a mirror of the battle-tested `spm_select`/`cfg_getfile` files they were copied from, to minimise the maintenance burden.
> [30526b9#diff-801283b7ca73698423622e9903c08949b16555ade75d6268a33a89b3e731bafeR155-L138](https://github.com/bids-standard/bids-matlab/commit/30526b9f53c69ad388faf50cc512941f9f3a40c0#diff-801283b7ca73698423622e9903c08949b16555ade75d6268a33a89b3e731bafeR155-L138)

Happy to revert, as long as we can make the code more readable: 
- more explicit variable names
- if possible make the control flow less of a headache to read

Ended up writing unit tests to make sure I would not be breaking stuff while refactoring.

If I struggled to make sense of them, I suspect others will. :-)

nbeliy:
I probably missing something here. 
You want to revert and make it unusable on windows?

Remi-Gau:
Ah sorry for not explaining.

The original code was taken from spm and is known to work on Windows.

The commit he highlighted was from some refactoring I did to the code and that is likely where the bug was introduced.

[...]

PR #166, opened 2021-02-22
Bug in file_utils

nbeliy:
Fixed bug described in #165 

It was caused in `canonicalise_path` during reconstruction of path. Linux path was assigned before the check for windows.
This PR fixes it.


Remi-Gau:
Thanks for that one!!!

Remi-Gau (approved):

PR #167, opened 2021-02-22
Small changes in anat entities

agahkarakuzu:
Following the latest [changes,](https://bids-specification.readthedocs.io/en/latest/99-appendices/10-file-collections.html) the `fa` entity will be changed to `flip` to avoid confusion with the FA of Fractional Anisotropy. 

A new entity for `MTState` metadata field `mt` is also introduced. 

I reflected these changes here, not sure if there is some other file I need to modify. 

Lastly, I don't see the `part` entity in the list. Is that defined elsewhere, or should I add it here as well? 

Thanks!


Remi-Gau:
Hey Agah: in theory all the hard coding of the entities will be removed when we merge the `dev` branch.

It regularly imports the latest version of the bids-schema and uses that to parse the filenames.

It should also provide a schema less parsing that should pick up any file of the shape `sub-label[_entity-label]_suffix.extension`.

I still need to do some testing on all the new stuff from BEP001 using the bids-examples.

Sorry a lot of that has been happening "behind the scenes". Things should get clearer soon. I hope.

agahkarakuzu:
That's great news @Remi-Gau ! 

I will deal with this from the fork, as I keep an eye on the developments here. Happy to see that the matlab version is up to speed! Thank you so much for the effort 🚀

I will close this PR.

PR #171, opened 2021-02-23
[WIP] Function to copy derivatives

### LINKED ISSUE #78: Initialize derivative data set with a copy of the raw data
There are a few circumstances (see here under) where I would like to initialize the `derivatives` with a copy, complete or partial, of the original full BIDS dataset. I have been working on a `prepare_derivatives.m` for that specific purpose with some features, such as:
- copy the data of all or a sub-list of subjects;
- if there are multiple sessions, copy the data from all or only some sessions;
- copy all or a sub-list of modalities, and for `func` data possibly all or some tasks;
- unzip the `.nii.gz` files and 4D-to-3D unpacking the `func` and `dwi` images, if requested;
- decide where that `derivatives/tool` subfolder will be created, i.e. not necessarily next to the "raw" data in the BIDS folder.

Why would I want to do that? Well because, it is convenient to work on a subset of data and/or in a "sandpit", for example
- to develop a processing pipeline with 5 subjects instead of 100 or simply to keep the subjects that are kept for the specific analysis, i.e. leave out some "outliers"
- analysis can focus on a single session for some question and therefore the processing will be different from that combing multiple sessions
- analysis can also focus on a single modality (or task for `func` data), therefore no need to bring along everything
- some operations in SPM's standard special processing pipelines, e.g. realignment and coregistration, do change the header of the images, so they ought to be in a `derivatives` subfolder before any processing. Moreover SPM and some other tools, would prefer dealing with a series of 3D images rather than a single 4D volume, e.g. ditching the 1st few functional images into a "dummy" subfolder.
- if the dataset is saved on an external storage place (like the centralized "mass-storage" system at ULiège), one cannot directly work on this server and therefore data need to be copied locally before any processing is applied, i.e. I might as well call this my `derivatives\some_tool_or_step` data and start working on these. Afterwards, it is straightforward to copy that derivatives folder back on the external storage place.

_Originally posted by @ChristophePhillips in https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-714505816_

- possibility to copy part of the raw only by filtering by subject, session, modality, task
- possibility to unzip and split 4D into 3D (FYI: Rémi is not in favor of this. :wink: )
- this derivative folder can be anywhere relative to the raw data but make the default path follow one of the recommended ways to store derivatives with respec to the raw (see [BIDS specs here](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#source-vs-raw-vs-derived-data)
- initialize a `dataset_description.json` in the root of that derivatiive folder

---

See comments:

- https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-714535229
- https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-717327122
- https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-717345065
- https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-717402112

--- 

In other repos:

- https://github.com/Remi-Gau/CPP_BIDS_SPM_pipeline/blob/bf850e944b0f9e2ee8381c11b2a3ef21960da54c/src/workflows/bidsCopyRawFolder.m
- https://github.com/CPernet/spmup/blob/master/bids/spmup_BIDS_unpack.m

Remi-Gau:
fixes #78

Added a function`derivate(BIDS, out_path, name, varargin)` which exports the files from current BIDS layout into `out_path/name` folder. In other words exports data into derivatives.

`varargin` are the query options same as in `bids.query`, in fact they are passed directly to it.

To do (**help needed!**):

- [x] Options -- when created derivatives, there a list of usefull options to fine tune, but adding them will make list of parameters to long. Options that I would like to include are below.
- [ ] ~~Script add a line to `dataset_description` a new structure each time it is run. But on the third run it crashes (see below)~~ referred into another PR.
- [ ] ~~Add version of bids-matlab (does it exists?)~~ TODO when when have a version number.
- [ ] ~~Optimise the loop on files~~ (see #161)
- [ ] ~~Copy of derivatives. Actually it supports only full paths in `dependencies.*`field.~~ Refered in another issue
- [ ] Copy of `participants.tsv` and `scans.tsv` files
- [x] Tests

### Options

- [x] *unzip* -- if true, then zipped files are unzipped
- [x] *copy\_dependencies* -- if true, the files that current file depend on are copied even if they are not explicitely queried
- [x] *force* -- copy file even if he already exists in output. Mutually exclusve with *copy\_dependencies* to avoid circular copy

### Crashes during `dataset_description` generation:
```
Error using horzcat
Dimensions of matrices being concatenated are not consistent.

Error in bids.derivate (line 68)
    description.GeneratedBy = [description.GeneratedBy pipeline];
```


Remi-Gau:
PR #162 reopened from the feature branch

Remi-Gau:
Once the tests check out green I will merge this into `dev` and open issues for the remaining `to dos`

Issue #157, opened 2021-02-18
Intended for does not cover qMRI cases


Remi-Gau:
Sorry to reopening the request. There two issues that I encountered during the tests:

1. Do not take into account qMRI fieldmaps (TB1EPI etc...)
2. The json list in IntendedFor is transformed into cell and not structure, hence it avoid  elseif on line 527
```
ans =

  6x1 cell array

    {'anat/sub-HC1278_echo-*_acq-T1w_part-phase_MPM.nii.gz'  }
    {'anat/sub-HC1278_echo-*_acq-T1w_part-mag_MPM.nii.gz'    }
    {'anat/sub-HC1278_echo-*_acq-MToff_part-mag_MPM.nii.gz'  }
    {'anat/sub-HC1278_echo-*_acq-MToff_part-phase_MPM.nii.gz'}
    {'anat/sub-HC1278_echo-*_acq-MTon_part-mag_MPM.nii.gz'   }
    {'anat/sub-HC1278_echo-*_acq-MTon_part-phase_MPM.nii.gz' }
```

Proposition of solution (not sure if it will broke something else):
1a) Search IntendedFor in each of json files
1b) Search IntendedFor in each file in modality fmap, not sure how it will work for coodsystem and m0

_Originally posted by @nbeliy in https://github.com/bids-standard/bids-matlab/issues/151#issuecomment-781324715_

Remi-Gau:
> 1b) Search IntendedFor in each file in modality fmap, not sure how it will work for coodsystem and m0

Ah thanks.
As the qMRI examples are not yet in the bids-examples repo I had overlooked this.

Yeah I think the best is to check all the files in fmap. Will fix this.

Remi-Gau:
> Yeah I think the best is to check all the files in fmap. Will fix this.

ha but if we do that then we end up with all the files in fmap being listed... OK need to think about this a bit.

Issue #143, opened 2021-02-15
generalize usage of `intended_for` field where applicable and `dependency` sub-structure


Remi-Gau:
- [ ] generalize usage of `intended_for` field where applicable and `dependency` sub-structure

_Originally posted by @Remi-Gau in https://github.com/bids-standard/bids-matlab/issues/124#issuecomment-779079965_

For the in the intended for, it should at least cover fmaps, m0scan and similar files that have an intended for in their metadata

For dependencies (tempted to call it `informed by` as pybids does it), it could be more flexible and could / should include events.tsv files for any "task" file for example.

Issue #130, opened 2021-02-05
Dealing with intendedFor and "file dependencies"


Remi-Gau:
> So I would not necessarily think this is needed for `events` but definitely, we need a way to deal with the `intendedFor` for fieldmaps, especially that it is being reused for other data types as well (ASL, PET...) .

@nbeliy 

The new PR for ASL seems to incorporate the idea you had for "file dependencies" as aspects of ASL includes the "intendedFor" field

See PR #127 and also the code here: https://github.com/ExploreASL/bids-matlab/blob/6cba2ee686fc26fb9b182f6a50a99ae15a0a11fa/%2Bbids/layout.m#L481

I would suggest that we start small and tackle first those "explicit" dependencies covered by the specification before we move on to those that are more use case dependent.

What do you think?

_Originally posted by @Remi-Gau in https://github.com/bids-standard/bids-matlab/issues/122#issuecomment-774021374_

Remi-Gau:
from @nbeliy

> I would suggest that we start small and tackle first those "explicit" dependencies covered by the specification before we move on to those that are more use case dependent.

In my idea dependency is to revert IntendedFor. A fmap image do not have a value by itself, and it is not that important to know for what scan it was taken. But for a given scan it is important to have an associated fmap. Actually it would be nice to have it in BIDS itself, but I guess is too late.

From technical point of view, it can be done at the end of layout function. Just scanning all images for IntendedFor field, and adding found files to corresponding scans.

Remi-Gau:
I agree that this makes more sense when using the data indeed.

@HenkMutsaerts what do you think?

Issue #146, opened 2021-02-15
allow layout to index files in the derivatives folder a raw dataset


Remi-Gau:
This is a valid BIDS directory structure, so it should be possible /convenient to index the content of some of those extra folders:

```
tests/bids-examples/ds000117/
├── derivatives
│   ├── freesurfer
│   └── meg_derivatives
├── stimuli
│   ├── func
│   └── meg
├── sub-01
├── sub-02
...
└── sub-emptyroom
```

Remi-Gau:
These files should also be potentially copied to a new derivatives when copying the raw data: see https://github.com/bids-standard/bids-matlab/pull/171#issue-578171109

PR #172, opened 2021-02-23
Improvements on dependencies calculations

Remi-Gau:
#### Metadata treatment
Split `get_metadata` into two functions:

- `metalist = get_meta_list(filename, pattern)` returns list of files satisfying pattern with inheritance principle
- `get_metadata(metafile)` construct meta structure from json files in passed meta-files list (from above function)

Each data file in dataset receives new field `metafile` that contains list of files from `get_meta_list`.

This will facilitate access to metafiles directly. Individual values can be extracted by  user if needed using `get_metadata`.

#### Dependencies

Each file structure contains dependencies sub-structure with guaranteed fields:

- `explicit` -- list of data files containing "IntendedFor" referencing current file.
- `data` -- list of files with same name but different extension. This will combine files that are split in header and data (like in Brainvision), also takes care of bval/bvec files
- `group` -- list of files that have same name except extension and suffix. This will group file that logically need each other, like functional mri and events tabular file. It also takes care of fmap magnitude1/2 and phasediff

Additional custom subfields can be added for given modalities.

This allow to treat `dwi` and `fmap` with generic `parse_using_schema`. `perf` likely also, but I can't certify it.

#### Issues

- [ ] performance. I feel that retrieval of dependencies can be optimised
- [x] Tests -- was unable to install moxunit
- [x] Tests on other datasets


Remi-Gau:
This should:
- fix #130 
- fix #143 
- partly fix #157 

For qMRI some files in the `derivatives` folder also have an `intendedFor` and they would be missed by the current code.

I suggest we finish this PR as for now and then tackle #146 that concerns the indexing of nested derivative folders before we come back to finish fixing #157 completely.

Remi-Gau:
>     * [ ]  performance. I feel that retrieval of dependencies can be optimised

I suspect that having to load and read the metadata of each file takes a lot of time.

I suspect that this can indeed be optimized but I would worry about that later.

Remi-Gau:
@nbeliy 

I pretty much took your code as is, fixed the test that were not passing and did a bit of refactoring and commenting.

I also:
- did a soft reset to ditch the `derivatives` function from this PR so we can tackle separately in your other PR
- cherry picked a commit that you had on your `copy derivatives` branch to update the query function

The code is pretty much the same though: thanks for providing all this.

Let me know what you think.

Note: I feel that the current way of indexing all the dependencies creates a lot of "redundancies" in the layout (example: the phase image of fieldmap will list all the associated imagesin `dependencies.group` like magnitude and the other way around). 

But I think that we can also optimize this later, if needed as this way of indexing things makes fewer assumptions about what file is / should be the "main" one for a given group.

nbeliy:
> For qMRI some files in the `derivatives` folder also have an `intendedFor` and they would be missed by the current code.

It will still work if `IntendedFor` in `derivatives` reference file in itself, not in the original/main dataset. If it's not, we can imagine a search for files in main dataset at condition that main dataset is referenced in dataset_description.json. But then, it will be difficult to implement, considering that main dataset may be defined as DOI or URL.

> I suspect that having to load and read the metadata of each file takes a lot of time.

Conserving full metadata for each file can explode memory usage. From inside we need only `IntendedFor`, and for user a query with `'target', {'meta1', 'meta2'}` can be imagined to retrieve several parameters with one go. 

> Note: I feel that the current way of indexing all the dependencies creates a lot of "redundancies" in the layout (example: the phase image of fieldmap will list all the associated imagesin dependencies.group like magnitude and the other way around).

Sometimes the main file is not well defined -- in phasediff and 2 magnitude, which one is main? In order to simplify we can exclude `.tsv` files from looking for group dependencies.

+bids/+internal/ends_with.m
```diff
@@ -0,0 +1,23 @@
+function res = ends_with(str, pattern)
+  %
+  % Checks id character array 'str' ends with 'pat'
+  %
+  % USAGE  res = bids.internal.endsWith(str, pat)
+  %
+  % str        - character array
+  % pat        - character array
```
nbeliy, line 8: Will it be a more complexity/performance issue to make it also accept cellstr for pat?

+bids/layout.m
```diff
@@ -182,159 +176,39 @@
...
-      subject = bids.internal.append_to_layout(file_list{i}, subject, modality, schema);
+            subject.(modality)(end).meta = bids.internal.get_metadata(subject.(modality)(iFile).metafile);
 
-      % if this file is a nifti image we add the bval and bvec as dependencies
-      if ~isempty(subject.(modality)) && ...
-              any(strcmp(subject.(modality)(end).ext, {'.nii', '.nii.gz'}))
+            aslcontext_file = strrep(subject.perf(end).filename, ...
+                                     ['_asl' subject.perf(end).ext], ...
+                                     '_aslcontext.tsv');
+            subject.(modality)(end).dependencies.context = manage_tsv( ...
```
nbeliy, line 198: Not sure that asl and aslcontext means, but if it only replaces suffix, will not the `_aslcontext.tsv` appear in group dependencies?

+bids/layout.m
```diff
@@ -452,114 +382,49 @@ function tolerant_message(use_schema, msg)
...
-
-    else
-      intended_for{end + 1, 1} = fullpath_filename;
-
-    end
   end
 
 end
 
 function perf = manage_asllabeling(perf, pth)
```
nbeliy, line 425: This `jpg` image also appears in `group` dependency

+bids/query.m
```diff
@@ -216,11 +219,16 @@
                 result{end + 1} = fullfile(BIDS.subjects(i).path, modalities{j}, d(k).filename);
               end
 
-            case 'metadata'
+            case 'metafiles'
               if isfield(d(k), 'filename')
+                fmeta = BIDS.subjects(i).(modalities{j})(k).metafile;
```
nbeliy, line 224: Don't see point of line 223

nbeliy (approved):
Looks good for mee, more abstract and generic.

As for optimisation, the list of candidates for metadata (considering only json) is same for each call. Probably it can be saved somewhere.

If (to be confirmed) data files are scanned in alphabetical order, lokking for group dependencies can be limited in file's neighbours.

Also, we probably need to move `get_metadata` from internal to util. This function can be usefull outside the layout and query

Remi-Gau:
Will try to adress some of your questions. Before merging.

Totally agree with getting get_metadata out of `internal`, was suggested before and I have been dragging my feet to do it. :man_shrugging: 

Was thinking something along the same line for optimization: will open a separate issue to keep track of that.

+bids/query.m
Remi-Gau: Yup you are correct some of those IF blocks are most likely to always turn out to be true.
Just removed them.

+bids/layout.m
Remi-Gau: correct

I removed the special way of handling this

+bids/+internal/ends_with.m
Remi-Gau: Should be easy enough: the idea would be to check if the file ends with any pattern from all the patterns in a cell?

Remi-Gau: FYI: Will most likely move this to a different issue.

+bids/layout.m
Remi-Gau: it will but it is also needed by `layout` to take care of the M0 so at the moment, it might be better to leave it there.

PR #178, opened 2021-02-26
[FIX] lower priority to use SPM function to deal with json files

### LINKED ISSUE #176: jsonwrite for outdated SPM
The interface of `spm_jsonwrite` changed in SPM in revision r. 6975, making it unusable for outdated SPM installations. 
Installation of JSONio do not change anything as the check for existence for `spm_jsonwrite` comes first.

I see two solution for this (other from keep SPM updated):

- use only JSONio, cite it as dependency
- Invert if in `+bids/+util/jsonencode.m` line 33, so JSONio takes priority

Tell me that you think

UPD:
You also check for native MATLAB jsonencode, but it has the lower priority. May be it can be rendered as highest one

Remi-Gau:
fixes #176 

- [x] change order of priority to check which function to use to deal with JSON: built-in functions > JSONio > SPM
- [x] reformat help sections of those functions

Once merged I will also apply this fix on the `dev` branch.

### DIFF
--- a/+bids/+util/jsonencode.m
+++ b/+bids/+util/jsonencode.m
@@ -1,20 +1,36 @@
 function varargout = jsonencode(varargin)
+  %
   % Encode data to JSON-formatted file
-  % FORMAT bids.util.jsonencode(filename,json)
-  % filename - JSON filename
-  % json     - JSON structure
   %
-  % FORMAT S = bids.util.jsonencode(json)
-  % json     - JSON structure
-  % S        - serialized JSON structure (string)
+  % USAGE::
   %
-  % FORMAT [...] = bids.util.jsonencode(...,opts)
-  % opts     - structure of optional parameters:
-  %              Indent: string to use for indentation [Default: '']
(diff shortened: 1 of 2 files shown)


+bids/+util/jsonencode.m
```diff
@@ -59,8 +76,21 @@
       fprintf(fid, '%s', txt);
```
nbeliy, line 76: Will it write to file with new lines and indend?

Remi-Gau: I just checked and if you rely on Matlab built-in function there will not be any JSON "formating".

I might be wrong but I don't think those options are supported by Matlab's functions. :sob:

gllmflndn: My little finger told me to tell you to wait a few weeks for `PrettyPrint`.

gllmflndn: https://www.mathworks.com/help/matlab/ref/jsonencode.html#bvb7dgc-2

Issue #118, opened 2021-01-25
JSON keys can't start with numbers


Remi-Gau:
from Mattermost. Not exactly due to bids-matlab code but a behavior that could become annoying for us.

> the bids-matlab json encode/decode functions can't handle json keys that start with numbers? This is for the simple reason that matlab struct field names can't start with numbers. I wonder if it's worth finding a workaround - this interferes with common use cases like defining a surface mesh density of, e.g., 32k, or an image resolution of 2mm.

gllmflndn:
You can use a [Map](https://www.mathworks.com/help/matlab/ref/containers.map.html) object instead of a struct (see [this](https://www.mathworks.com/help/matlab/ref/jsonencode.html#bvf4q2l-6)).
```
>> m.data = containers.Map({'32k','2mm'},{'mesh density','image resolution'})

m = 

  struct with fields:

    data: [2x1 containers.Map]

>> bids.util.jsonencode(m)                                                 

ans =

    '{"data":{"2mm":"image resolution","32k":"mesh density"}}'
```
You can also directly use a struct in Octave:

```
octave> m = struct('2mm','image resolution','32k','mesh density')
m =

  scalar structure containing the fields:

    2mm = image resolution
    32k = mesh density

octave> bids.util.jsonencode(m)
ans = {"2mm":"image resolution","32k":"mesh density"}
```

Decoding will be a bit more problematic as [matlab.lang.makeValidName](https://www.mathworks.com/help/matlab/ref/matlab.lang.makevalidname.html) will be used to construct _valid_ keys.

What are the actual situations where this issue arises? In the example above, `32k` and `2mm` look more like values than keys.

Remi-Gau:
Will add the relevant parts of the mattermost discussion here: (for reference).

> If you look at https://bids-specification.readthedocs.io/en/stable/05-derivatives/03-imaging.html#preprocessed-coregistered-andor-resampled-volumes, the Resolution field may be an object indexed by the values of the res- entity (similarly Density and den-). If somebody uses res-1 or  den-32k, you're going to get these.

> Found this: https://www.mathworks.com/matlabcentral/answers/435390-json-decode-lost-information

> Seeing from other discussions that people are complaining about losing order information on parse. IIRC matlab should give you access to Java objects. Java has an insertion-ordered hash map (https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html), in case somebody feels the urge to write a new parser.
[...]

PR #179, opened 2021-02-26
[FIX] default pattern of get_metada expects suffixes to be preceded by underscore

### LINKED ISSUE #175: Issues with bids.internal.get_metadata for derivatives
I'm encountering issues using bids.internal.get_metadata to extract metadata for derivative files. Specifically, I have a directory of Freesurfer-derived surface files, each with their own json file, including:

sub-01_hemi-R_space-fsnative_den-native_thickness.shape.gii
sub-01_space-fsnative_den-native_thickness.dscalar.nii
sub-01_hemi-R_space-fsnative_den-native_midthickness.surf.gii

The first file is a GIFTI file specifying cortical thickness; the second file is a CIFTI file combining cortical thickness measures from left and right hemispheres; the third is a representation of the right cortical surface (mid-thickness).

When I use bids.internal.get_metadata (default pattern) on the first file, information from the json metadata for files 2 and 3 is included, as if all of these json files apply to the first file.

The file sub-01_space-fsnative_den-native_thickness.json is included because its key/value pairs are a subset of the first file's. So get_metadata is properly applying the bids standard here, but the standard behaves weirdly in this case. Not sure how to address this.

The file sub-01_hemi-R_space-fsnative_den-native_midthickness.json is included because the file matches the pattern ^.*thickness\\.json$, and its key-value pairs are the same as the first file's. Here, the bids standard is not being properly applied - the midthickness json data should not be included.

Remi-Gau:
fixes #175 

- [x] change default pattern for get_metadata
- [x] add tests to check that "similar" suffixes are distinguished (like '_thicknes' and '_midthickness')
- [x] add test for the rare cases of files with no underscores (like `participants.tsv`)

The vast majority of files with metadata in BIDS have an underscore so making this the default seems acceptable to me, though this could be problematic when there is a single `events.tsv` file in the root directory that applies to all subjects, sessions by inheritance principle.

Will push to `dev` once this OK on master.


Remi-Gau:
Will keep around the test data added in this PR for later as they could be reused for the issue of  json field starting with a number.

See issue #118.

Thanks again @bmdeen for providing those.

Issue #177, opened 2021-02-26
Feature request: function to add new fields to `dataset_description.json`


nbeliy:
When creating derivatives, BIDS requires to add `GeneratedBy` field. On itself it is not difficult, but it would nice to have a simple centralised script to generate such fields. 

Something like:
```
update_description(BIDS, fieldname, data, append)
```
with append a bool that decides if field will be appended to existing field, or will replace it.

Remi-Gau:
Very mcuh agreed.

This is part of creating / enhancing the "writing" aspects of bids-matlab.

This issue is subset of some the feature suggestions from  @bmdeen in #79.

Will also have start having a look into this, as it could help create some utilities that could be useful for other things later.

Issue #79, opened 2020-10-29
Improve handling of the metadata


Remi-Gau:
A few basic functions that I would find useful, from the perspective of creating pipelines that generate/process derivatives in a BIDS-compliant manner:

- [x] A function for changing a BIDS filename, based on changes to specific key/value pairs (i.e., changing, adding, and removing key/value pairs to a name, and maintaining the proper order of fields in the name).
- [ ] A function for changing one or several fields in a .json metadata file (e.g., to generate a new metadata file after applying some transformation / preproc step to a dataset)
- [x] A non-private version of bids-matlab's get_metadata function. I frequently find myself wanting a simple function to check a specific metadata field of a single file, without having to run both bids.layout and bids.query. I understand the rationale for keeping get_metadata private, but from my perspective it would be very useful to have a public version of something like this command.

_Originally posted by @bmdeen in https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-716181111_

See also comments:

https://github.com/bids-standard/bids-matlab/issues/60#issuecomment-717327122

Remi-Gau:
Originally posted by @gllmflndn 

>@bmdeen: good point about get_metadata: we should have a user-level helper function to query the metadata of a given file.

Remi-Gau:
> Add a function for changing one or several fields in a .json metadata file (e.g., to generate a new metadata file after applying some transformation / preproc step to a dataset)

PR #184, opened 2021-02-28
[ENH] add basic function to generate and update dataset_description

Remi-Gau:
- [x] allow easy creation of dataset_description from scratch
- [x] allow update of existing dataset_description
- [x] allow to generate derivative description from source description
- [x] update copy_to_derivatives to use this class to update the dataset_description

---
In a separate PR

- [ ] ~~transfer some of dataset_description function from layout.m to this class~~

---

Something similar could be implemented to deal with metadata: this could help with some of the points from #79


Remi-Gau:
now fixes #177

PR #185, opened 2021-03-01
improved data files regexp

nbeliy:
Affined the file search regexp, should not conflict with anything.

Reason: SPM sometimes not only add a prefix, but also a suffix for temporary files. And the original queries accepts such files.
The original suffix is treated as entity, and crashes when tries to split by '-'

In layout:

  - requesting strart with `sub-<subId>_ses-<sesId>_`
  - requesting a set of entities of form `<key>-<value>_`
  - requestin exactly one suffix

In parse_filename

  - Defining prefix as everything coming before `sub-`
  - Checking that only one `sub-` is present

### DIFF
--- a/+bids/+internal/parse_filename.m
+++ b/+bids/+internal/parse_filename.m
@@ -39,8 +39,18 @@ function p = parse_filename(filename, fields)
 
   filename = bids.internal.file_utils(filename, 'filename');
 
+  % -Identify prefix as string coming before 'sub-'
+  pos = strfind(filename, 'sub-');
+  if numel(pos) ~= 1
+    warning(['File name ' p.filename ' not bids compatible']);
+    p = struct([]);
+    return
+  end
+  p.prefix = filename(1:pos-1);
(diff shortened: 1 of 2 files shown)


Remi-Gau:
Trying this in a separate PR.

> Affined the file search regexp, should not conflict with anything.
> 
> Reason: SPM sometimes not only add a prefix, but also a suffix for temporary files. And the original queries accepts such files.
> The original suffix is treated as entity, and crashes when tries to split by '-'
> 
> In layout:
> 
>     * requesting strart with `sub-<subId>_ses-<sesId>_`
>     * requesting a set of entities of form `<key>-<value>_`
>     * requestin exactly one suffix

> In parse_filename
> 
>     * Defining prefix as everything coming before `sub-`
>     * Checking that only one `sub-` is present

I will remove the constraints that filenames to parse must start with `sub-` as it was breaking other things and potentially prevent to parse group level files in future derivatives where there is no `sub` entitiy.

PR #187, opened 2021-03-04
Added check for data consistency

nbeliy:
Added new function to util with main goal to check if data returned from different bids.query are consistent, i.e. has same name (up to suffix).

Probably not the most optimal way to implement.

- [x] tests
- [ ] documentation (have no idea how encode option as switches)

### DIFF
--- a/+bids/+util/check_data_consistency.m
+++ b/+bids/+util/check_data_consistency.m
@@ -1,0 +1,123 @@
+function check_data_consistency(to_check, varargin)
+  %
+  % Given a set of filepaths cellarrays from bids.query, checks that any
+  % all paths at same index have consistent bids name and same size
+  % Raise error if inconsistency found
+  %
+  % Options:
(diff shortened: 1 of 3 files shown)


Remi-Gau:
hey,
I was trying to figure out what the use case for this would be.
Can you tell me a bit more?

nbeliy:
The queries in bids-layout are independent, but often will be used in correlation with another. And it will be critical that the reults are in the same order and same dimensions. 

For example first query is list of fMRI images, second query is list of task events. Both lists must have same name (except suffix and extension). So this function is just do this check.

Ideally it can be extended for cases check if all entities coincides except sub (for example), to check that 2 subjects have exactly same set of data.

nbeliy:
Hi Remi, I'm cleaning my branches, so I close this pull request, as it clearly outdated.

Remi-Gau:
okidoki

PR #207, opened 2021-04-18
[FIX] fix bugs in copy to derivatives

Remi-Gau:


nbeliy:
- Using unzip directly than copiyng file, unzip it and then removing
- Need to change query to filter in documentations

 I started to question the utility of copying files with dependencies.
Usually it is quite easy to retrieve the dependencies by adapting query.

Remi-Gau:
So far I kind of like the possibility of grabbing all the related files at once.

But there is the possibility to skip copying the dependencies:

https://github.com/bids-standard/bids-matlab/blob/f9beef23f2f9e94ca605507528d23c92a70a599f/tests/test_copy_to_derivative.m#L70

Remi-Gau:
- [ ] implement a way to "force" copy without getting caught into an infinite loop

The way dependencies are indexed, they reference each other ( `fileA_events.tsv` <--> `fileA_bold.nii`).

When using "force copy" and "do not skip dependencies", this leads to an infinite loop of copying.

At the moment I circumvented the problem by preventing to copy dependencies of dependencies: but this might be suboptimal as when copying func file, it might not copy all the necessary `fmap` for example.

Issue #97, opened 2020-11-13
Fail to capture all EEG files and wrong number of subjects


arnodelorme:
Only capture the raw data, channel and event file. Misses the JSON files.
Also, there are only 2 subjects (with 3 sessions each). Yet the program think there are 6 subjects

<img width="455" alt="Screen Shot 2020-11-13 at 10 49 21 AM" src="https://user-images.githubusercontent.com/1872705/99119934-1f974580-259e-11eb-9a0b-cc03bd547646.png">
<img width="521" alt="Screen Shot 2020-11-13 at 10 49 12 AM" src="https://user-images.githubusercontent.com/1872705/99119942-21f99f80-259e-11eb-94cf-f36aab20870c.png">

Remi-Gau:
hey @arnodelorme 

Try to use `bids.query` to access specific pieces of information of the layout. This is how the whole thing has been designed. :wink: 

I still need to write some sort of official doc on this but in the mean time you can look at [test_bids_query](https://github.com/bids-standard/bids-matlab/blob/master/tests/test_bids_query.m) as a source of unofficial doc.

In your case, what do the following give you? 

```
bids.query(res, 'subjects')
bids.query(res, 'data', 'sub', '01')
```

Also I am pretty sure that JSON files are not "indexed". Am I right @gllmflndn?

If you want to get the content of some JSON file you would need to do something like this:

```
bids.query(res, 'metadata', 'sub', '01', 'task', 'Experiment')
```

arnodelorme:
Why not load it all?

Remi-Gau:
Not sure why that initial decision was made though I suspect that this in part related to dealing with the inheritance principle which implies that the metadata about a given file could be spread in several JSON, so this is dealt with on the fly when metadata is requested.

Remi-Gau:
hey @arnodelorme

PR #47 should be able to index the json files that are in any "modality" folder a subject if you ask for a schema-less layout.

```matlab
use_schema = true();
BIDS = bids.layout(fullfile(pth_bids_example, 'eeg_rishikesh'), use_schema);
bids.query(BIDS, 'data', 'sub', '001', 'ses', '01')

ans =

  3×1 cell array

    'bids-examples/eeg_rishikesh/sub-001/ses-01/eeg/sub-001_ses-01_task-meditation_channels.tsv'
[...]

PR #243, opened 2021-05-31
Speed up dependencies indexing

### LINKED ISSUE #209: Inefficient dependencies calculations
The current way the dependencies are calculated are quite inefficient and may produce large lags on big datasets.

In my fork [nbeliy/bids-matlab/tree/beliy-dependencies](https://github.com/nbeliy/bids-matlab/tree/beliy-dependencies) I implemented a major optimization, which improve calculation by order of magnitude (seconds instead of minutes).
Unfortunately this fork diverged too much from the main dev, so I can't make quickly a new branch and merge request.

Idea is remember the index of latest group change (when the name up to suffix change) and basename change (name up to the extension).

If for a new file the basename is the same as in previous, all calculations will give the same result as in previous calculations, so we can just copy previous structure and adapt the filename and suffix. All such files will be in `data` dependencies, so we can go from to index where the basename changed for the last time and add current file to dependency, and for the current file, we can add only the previous one as dependency.

For the groups, is the same but without copy.

Modifications starts [there](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L182). 

**Obviously it relies that list of files is ordered alphabetically.**

- [L182:](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L182) initialisation of variables
```matlab
    index_group = 0;
    group_base = '';
    group_len = 1;
    index_data = 0;
    data_base = '';
    data_len = 1;
```

- [L192:](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L192) If basename does not change, just copiyng previous structure with change of filename, extension and dependency
```matlab
      if strncmp(data_base, file_list{i}, data_len)
        subject.(modality)(end+1, 1) = subject.(modality)(end, 1);
        subject.(modality)(end, 1).ext = file_list{i}(data_len:end);
        subject.(modality)(end, 1).filename = file_list{i};
        dep_fname = fullfile(pth, subject.(modality)(end-1, 1).filename);
        subject.(modality)(end).dependencies.data{end+1} = dep_fname;
```

- [L210:](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L210) looping over files with same basename and adding current file to dependency
```matlab
        for di = index_data:i-1
            subject.(modality)(di).dependencies.data{end+1} = fullpath_filename;
          end;
```

- [L215](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L215) basename changes, setting new values
```matlab
         index_data = i;
          data_len = find(file_list{i} == '.', 1);
          data_base = file_list{i}(1:data_len);
```

- [L219](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L219) Looping over files in the same group and adding the dependencies
```matlab
          for gi = index_group:i
            dep_fname = fullfile(pth, subject.(modality)(gi).filename);
            subject.(modality)(end).dependencies.group{end+1} = dep_fname;
            subject.(modality)(gi).dependencies.group{end+1} = fullpath_filename;
          end
```

- [L228](https://github.com/nbeliy/bids-matlab/blob/64fd39e42caeb5c10ed6b7e23e85a3e5f7fc6ad5/%2Bbids/layout.m#L228) Group changes, setting new values
```matlab
        index_group = i;
        group_len = find(file_list{i} == '_', 1, 'last');
        group_base = file_list{i}(1:group_len);
        
        index_data = i;
        data_len = find(file_list{i} == '.', 1);
        data_base = file_list{i}(1:data_len);
```

Remi-Gau:
fixes #209 
supersedes #185

- [x] speed up indexing by copying dependencies from the previously indexed file when possible / appropriate
- [x] improve regexp on filename and prefix parsing
- [x] minor changes in `copy_to_derivatives`
  - [x] unzip files directly when possible
  - [x] make pipeline name an obligatory argument
  - [x] change default ouput path


+bids/layout.m
```diff
@@ -365,36 +378,56 @@
...
-  subject.(modality)(end).metafile = bids.internal.get_meta_list(fullpath_filename);
-  subject.(modality)(end).dependencies.explicit = {};
-  subject.(modality)(end).dependencies.data = {};
-  subject.(modality)(end).dependencies.group = {};
+  if ~isfield(subject.(modality)(end), 'dependencies') || ...
+          isempty(subject.(modality)(end).dependencies)
+    subject.(modality)(end).dependencies.explicit = {};
+    subject.(modality)(end).dependencies.data = {};
+    subject.(modality)(end).dependencies.group = {};
+  end
```
nbeliy, line 386: The creation of dependencies field should be defined where the structure `subject.(modality)(end)` is created (i.e. in `append_to_layout` function). It will assure that subjects structure remains consistent.

+bids/layout.m
```diff
@@ -214,6 +214,15 @@
...
+                            'index_group', 0, ...
+                            'group_base',  '', ...
+                            'group_len', 1, ...
+                            'index_data', 0, ...
+                            'data_base', '', ...
+                            'data_len', 1);
+
     for iFile = 1:size(file_list, 1)
 
       [subject, parsing] = bids.internal.append_to_layout(file_list{iFile}, ...
```
nbeliy, line 228: It will be more efficient to not append to layout files with same base name but just copying the previous object.
This will allow to evade the recalculations of dependencies for the same file.

+bids/layout.m
```diff
@@ -365,36 +378,56 @@
...
-  for ii = 1:numel(candidates)
+    %     subject.(modality)(end + 1, 1) = subject.(modality)(end, 1);
+    %     subject.(modality)(end, 1).ext = file(dep_prev_files.data_len:end);
+    %     subject.(modality)(end, 1).filename = file;
 
-    if strcmp(candidates{ii}, file)
-      continue
-    end
+    dep_fname = fullfile(pth, subject.(modality)(end - 1, 1).filename);
+    subject.(modality)(end).dependencies.data{end + 1} = dep_fname;
```
nbeliy, line 395: This will work only if last subject is clone of subject before it -- see first comment.

Overwise a for loop is needed starting from `index_data` to `end - 1`

nbeliy: Same remark for metafile field. If last file is not clone, you need to copy the metafile attribute.

nbeliy (commented):
It should work. The only problem that you don't clone the last file is it is with same base name.
So in this case you need to update metafile (just copying it from previous file?) and already calculated dependencies.

nbeliy:
For some reason github don't allow me to see changes if I login. I'll put review in comments.

`append_to_layout.m`:
l31: Should we put a warning here? To tell user that some files are skipped?
l74: Just in case bids will add more formats that are folders, it will be more generic to test if path is directory.

`copy_to_derivatives.m`:
l3: Misleading description, in implementation function returns nothing. Bu it will be good to return the path to created derivative folder (e.g. `derivatives/bids-matlab`)
l24: To my taste, the pipeline name argument must be mandatory.
l94: From my understanding of (bids)[https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#storage-of-derived-datasets], by default derivatives should be within dataset folder, so `out_path = fullfile(BIDS.pth, 'derivatives');`. Also you provide default value to `out_path` -- it will be never empty so if will never trigger.
l258: This will be slow. You first copy data, then extract it, and then remove zipped files. You can extract files directly to destination. For ex. in `copy_with_symlink` add `if ends_with('.gz') && unzip; unzip_data(...)`. Obviously in this case you need to drop of removal of original zipped files.

`layout.m`:
It takes ~4s per subject (each containing ~800 files), so performance is improved.
However it treats json files as data, the search patterns must be adapted, for ex. like in #185 

Overwise, it looks good for me

Remi-Gau:
> `layout.m`:
> It takes ~4s per subject (each containing ~800 files), so performance is improved.

:smile: 

> However it treats json files as data, the search patterns must be adapted, for ex. like in #185

Yup I am working on it on a different PR in large swath inspired by #185.

Remi-Gau:
> `append_to_layout.m`:
> l31: Should we put a warning here? To tell user that some files are skipped?
> l74: Just in case bids will add more formats that are folders, it will be more generic to test if path is directory.

Done

nbeliy:
Tell me when I can retest

Remi-Gau:
> l24: To my taste, the pipeline name argument must be mandatory.

Agreed. Done. Mind you this change the order of input arguments.

> l94: From my understanding of (bids)[https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#storage-of-derived-datasets], by default derivatives should be within dataset folder, so `out_path = fullfile(BIDS.pth, 'derivatives');`. Also you provide default value to `out_path` -- it will be never empty so if will never trigger.

Good point. Actually the whole thing about folder organizations of derivatives and raw are mostly suggestions, but I agree that we can make the default to nest the derivatives in the raw.

> l258: This will be slow. You first copy data, then extract it, and then remove zipped files. You can extract files directly to destination. For ex. in `copy_with_symlink` add `if ends_with('.gz') && unzip; unzip_data(...)`. Obviously in this case you need to drop of removal of original zipped files.

Yup I know but I am pretty sure (as far as I can recall) I had to do things that way because the behavior of Octave and Matlab is slightly different with gunzip. Need to double check.

nbeliy:
> Yup I know but I am pretty sure (as far as I can recall) I had to do things that way because the behavior of Octave and Matlab is slightly different with gunzip. Need to double check.

From quick look into docs, the only difference I could spot is that matlab automatically creates folder if it not exists. But in any way, this has no relations with  dependencies calculations.

Remi-Gau:
latest commits implements #185 but relaxes constraints to have more than one instance of the string `sub-` in the filename: we'll just assume that the prefix is what comes before the first one.

Remi-Gau:
So the difference in gunzip behavior (and not `unzip` as the name of function implies) between Octave and Matlab is that Octave deletes the source file.

Here is MWE below. 

Will try to add a work around with an `if is_octave` to keep the speed up with matlab.

Not sure how if `gunzip` plays nice with symbolic links when working with datalad datasets. Will try it out.

### Script

```matlab
src_folder = fullfile(pwd, 'input');
tgt_folder = fullfile(pwd, 'output');

% set up
copyfile('sub-01_T1w.nii.gz', src_folder);

%do
gunzip(fullfile(src_folder, 'sub-01_T1w.nii.gz'), tgt_folder)
```

### Starting point

```bash
├── gunzip_cmp_matlab_octave.m
├── input
├── output
└── sub-01_T1w.nii.gz
```

### Results

#### With Matlab

```bash
├── gunzip_cmp_matlab_octave.m
├── input
│   └── sub-01_T1w.nii.gz
├── output
│   └── sub-01_T1w.nii
└── sub-01_T1w.nii.gz
```

#### With octave

```bash
├── gunzip_cmp_matlab_octave.m
├── input
├── output
│   └── sub-01_T1w.nii
└── sub-01_T1w.nii.gz
```

Remi-Gau:
> Tell me when I can retest

@nbeliy 

It should be good for another cross check.

Added a test for copy derivatives and unzip to try to make sure Octave and Matlab behave the same way on Linux.

Will quickly run things on Windows.

+bids/copy_to_derivative.m
```diff
@@ -285,39 +282,58 @@ function copy_with_symlink(src, target, verbose)
...
+
     if status > 0
       msg = ['Copying data with system command failed: ' ...
              'Will use matlab/octave copyfile command instead.\n', ...
              'May be an issue if your data set contains symbolic links' ...
              '(e.g. if you use datalad or git-annex.)'];
       bids.internal.error_handling(mfilename, 'copyError', msg, true, verbose);
-      use_copyfile(src, target, verbose);
+
+      use_copyfile(src, target, unzip_files, verbose);
```
nbeliy, line 306: I doubt that if the system copy fails, the matlab one will work

+bids/copy_to_derivative.m
```diff
@@ -285,39 +282,58 @@ function copy_with_symlink(src, target, verbose)
   % Though datalad should run on windows too
   %
 
-  command = 'cp -R -L -f';
-
   if verbose
     fprintf(1, '\n copying %s --> %s', src, target);
   end
 
   if  isunix
```
nbeliy, line 289: Links and datalad can be used on mac

+bids/copy_to_derivative.m
```diff
@@ -285,39 +282,58 @@ function copy_with_symlink(src, target, verbose)
...
              '(e.g. if you use datalad or git-annex.)'];
       bids.internal.error_handling(mfilename, 'copyError', msg, true, verbose);
-      use_copyfile(src, target, verbose);
+
+      use_copyfile(src, target, unzip_files, verbose);
     end
 
   else
-    use_copyfile(src, target, verbose);
+    use_copyfile(src, target, unzip_files, verbose);
```
nbeliy, line 310: Just to be consistent, I would do something like:
```
elseif ispc
  use_copyfile
else
  print warning(unknown system, the copy may fail)
  use_copyfile
end
```

nbeliy (approved):
Looks good for me. Performance on my tests is the same:
```
>> tic; BIDS = bids.layout('funct_3D', false); toc;
Elapsed time is 13.879293 seconds.
>> size(bids.query(BIDS, 'data')) 
ans =
        4793           1
```
It is still a waiting time but I'm impression we hit matlab itself.

I checked the `Approuve`, but for some reason ity didn't worked. For me it works and I "approve" the merge :-D

nbeliy:
P.S. How is the advances on regexp for file selection? I can't find the corresponding merge request

Remi-Gau:
> P.S. How is the advances on regexp for file selection? I can't find the corresponding merge request

Sorry.

Was done on local branch that I merged here:

https://github.com/bids-standard/bids-matlab/pull/243/commits/e43f2e374bbd5a9aec236f91756e944eb7fa72d7

nbeliy:
> Was done on local branch that I merged here:

Ok thet's why I still have json -- you exclude them only for datasets following the schema.
Is there reason why shemaless datasets should have json files as data?

Remi-Gau:
> Is there reason why shemaless datasets should have json files as data?

It was suggested by @arnodelorme in here
https://github.com/bids-standard/bids-matlab/issues/97#issue-742782287

nbeliy:
> It was suggested by @arnodelorme in here

Then will it be possible to get json files only if they are not metadada for data?
A json file will be created for each file in derivative, then to do query I will need to add `'extention', '.nii'` each time. 

Also, if he want to get metadata json, they can be retrived with `query('metafiles')`

Remi-Gau:
> > It was suggested by @arnodelorme in here
> 
> Then will it be possible to get json files only if they are not metadada for data?
> A json file will be created for each file in derivative, then to do query I will need to add `'extention', '.nii'` each time.
> 
> Also, if he want to get metadata json, they can be retrived with `query('metafiles')`

Would you be ok if I opened a separate issue for this: I feel that otherwise we are ending up changing behavior on this PR that was supposed to be only about optimization and has already devolved into way more than this.

Tempted to have an issue + meeting where several of the interested people on this can come together an agree on what we would like the API and the behavior to look like for a first version.

+bids/copy_to_derivative.m
Remi-Gau: good point

nbeliy:
> Tempted to have an issue + meeting where several of the interested people on this can come together an agree on what we would like the API and the behavior to look like for a first version.

It could be a good idea, if I will be invited)

Remi-Gau:
> > Tempted to have an issue + meeting where several of the interested people on this can come together an agree on what we would like the API and the behavior to look like for a first version.
> 
> It could be a good idea, if I will be invited)

most definitely

+bids/copy_to_derivative.m
Remi-Gau: implementing this now and testing: will merge with no further review once all test spass

+bids/copy_to_derivative.m
Remi-Gau: implementing this now and testing: will merge with no further review once all test spass

PR #252, opened 2021-06-15
Fixed bug of copying_to_derivative using structure

nbeliy:
There was a small bug in the argument parsing of `bids.layout`.

The check if `root` is structure was performed only if number of input arguments is 1, 
calling with optional parameter `use_schema` (like it is used in `copy_to_derivative`) 
skips this check and crashes on line
```matlab
if ~exist(root, 'dir')
```

This PR fixes it.


nbeliy:
Please disregard. Was using outdated dev branch

Issue #267, opened 2021-09-13
bids layout craches when processing bids-incompatible filenames


nbeliy:
When `bids.layout` encounters badly formatted bidsified names, the internal function `bids.internal.parce_filename` returns 
an empty structure, which usually crashes the following code (it tries to access `suffix`, `entities` fields). 

Crash is not followed by any meaningfull error/warning message, because `verbose` option in `parce_filenames` is never used.

Examples of problematic names:
```
func_task-AXcpt_bold.json
anat_T1w.json
task-AXcpt.json
```

Remi-Gau:
your issue and PR are super timely: I had literally made a note last week to tackle those issues, I had encountered super recently too.

PR #268, opened 2021-09-13
Added treatment of bids-incomatible files in dataset

nbeliy:
This pull request addresses issue #267 

Main changes:

- Removed `verbose` option from `parce_filename`, now warnings are always shown
- Slightly changed entities treatment:
  1. removing extention
  2. splitting basename into entities parts
  3. splitting each part into key-value
  4. based on number of keys-values pairs determining if it's suffix or entity
- based on above, providing more explicit warning message
- After `parse_filename` is called, checking if returned structure is empty
- Adding `'tolerant', false` where it make sense

Standing issues:

- if malformed files are json ones, they are rechecked regularly, spamming terminal output


Remi-Gau:
General question:

What SHOULD be the expected behavior when of `parse_filename` when using it in `tolerant` mode on a bids invalid file?

Would it make sense for `parse_filename` to at least return `p.filename = filename` in some casess?

The reason I am wondering about it is also to because what should be the expected behavior of the following?

```
tolerant = true;
use_schema = false;
index_derivatives = false;
BIDS = layout(path_to_bids, use_schema, index_derivatives, tolerant);

data = bids.query(BIDS, 'data')
```

My assumption is that we would want this to list all the files that are in the dataset no matter if they are bids valid or not?

Remi-Gau:
Will try fixing the tests first and add some more to cover the blind spots that this PR adresses

nbeliy:
> What SHOULD be the expected behavior when of `parse_filename` when using it in `tolerant` mode on a bids invalid file?

I like current state: tolerant switches from error to warning, and the function calling `parse_filename` decide how to act on tolerated files. My initial goal is just report the files that failed instead of just random non descript error. For strict bids datasets it should produce errors, but for non strict, these files can be just ignored (but still reported) 

> Would it make sense for `parse_filename` to at least return `p.filename = filename` in some casess?

Not sure, the `[]` clearly indicates failed state, if the return will contain partially filled structure, the test for failed state will be more intricated. the function that calls `parse_filename` knows already the file name to parse

> The reason I am wondering about it is also to because what should be the expected behavior of the following?

I'll say it should put in dataset maximum files that still follow bids syntax. Just makes it easier for unregulated bids datasets.

Remi-Gau:
> > What SHOULD be the expected behavior when of `parse_filename` when using it in `tolerant` mode on a bids invalid file?
> 
> I like current state: tolerant switches from error to warning, and the function calling `parse_filename` decide how to act on tolerated files. My initial goal is just report the files that failed instead of just random non descript error. For strict bids datasets it should produce errors, but for non strict, these files can be just ignored (but still reported)

Yeah I got the same hunch, just wanted to cross check with you.
 
> > Would it make sense for `parse_filename` to at least return `p.filename = filename` in some cases?
> 
> Not sure, the `[]` clearly indicates failed state, if the return will contain partially filled structure, the test for failed state will be more intricated. the function that calls `parse_filename` knows already the file name to parse

yeah that makes sense: parse_filename does enough already, so let's leave it at that

> > The reason I am wondering about it is also to because what should be the expected behavior of the following?
> 
> I'll say it should put in dataset maximum files that still follow bids syntax. Just makes it easier for unregulated bids datasets.

Will cross check how pybids handles those cases: I would not mind if the general behavior was fairly similar across tools.

Remi-Gau:
>     * Adding `'tolerant', false` where it make sense

For now, I have removed the ones you had added that were in `create_filename`, `create_path` and `derivatives_json`

+bids/+internal/parse_filename.m
```diff
@@ -78,39 +76,81 @@
...
+  for i = 1:numel(parts)
+
     try
-      [d, dummy] = regexp(parts{i}, '(?:\-)+', 'split', 'match'); %#ok<ASGLU>
-      p.entities.(d{1}) = d{2};
-    catch
-      msg = sprintf(['Entity-label pair %s of file %s is not valid.\n', ...
-                     'This could also be to a suffix with an extra _'], parts{i}, filename);
+
+      if isempty(parts{i})
```
Remi-Gau, line 95: am struggling to come up with a test case that hits this. Can you think of one?

Remi-Gau:
> > > The reason I am wondering about it is also to because what should be the expected behavior of the following?
> > 
> > 
> > I'll say it should put in dataset maximum files that still follow bids syntax. Just makes it easier for unregulated bids datasets.
> 
> Will cross check how pybids handles those cases: I would not mind if the general behavior was fairly similar across tools.

Checked with my fellow bids maintainers and in brief pybids has an extra "validate" flag (lightweight filename based validation similar to what is done by parse_filename): if this is set to "false", then even bids invalid files will be indexed.

I think that implementing this behavior could be done but it is beyond the scope of this PR and I suspect it might be potentially annoying enough to implement: I suggest we wait till we have users requests for it.

+bids/+internal/parse_filename.m
Remi-Gau: FYI I am ok to leave it in but I might add a comment to mention that this is unlikely to ever happen.

nbeliy: Just to cover weird cases with double '_':
```matlab
[parts, dummy] = regexp('_a_', '(?:_)+', 'split', 'match')

parts =

  1x3 cell array

    {0x0 char}    {'a'}    {0x0 char}
```

nbeliy: Actually, it just for cases of starting and ending with '_', it is nt catching duble underscore:
```matlab
parts, dummy] = regexp('a__b', '(?:_)+', 'split', 'match')

parts =

  1x2 cell array

    {'a'}    {'b'}


dummy =

  1x1 cell array

    {'__'}
```

Remi-Gau: Cool!!! 

Added a test to cover that error too.

Remi-Gau:
Merging this now! Thanks @nbeliy

Remi-Gau (approved):

Issue #282, opened 2021-11-03
Counterintuitive bids.File interface


nbeliy:
I find the new interface to `bids.File`, trying to mix structure and class fields.
For example function `create_filename` don't create a file name, but makes a copy of initial object with some changed fields.
As using objects, clone objects is trivial, I propose to instead of functions `create_filename`, `create_rel_path` etc. use one function `update()` that based on values of internal fields changes filename, relative path etc.

In that way, user can parse a file, change fields directly and regenerate new name in transparent way:
```matlab
file = bids.File('sub-AD0041_desc-sum_rangeLow-9_rangeHigh-22_pet.nii');
file.prefix = 'abcde';
file.update();
file.filename

  ans = 'abcdesub-AD0041_desc-sum_rangeLow-9_rangeHigh-22_pet.nii'
```
Should I make a merge request for that?

Remi-Gau:
Yes that you would make more sense indeed. Open a PR and we will take it from there.

Remi-Gau:
closed by #289

PR #289, opened 2021-11-08
Improvement of File.m interface

nbeliy:
This pull implements refactoring to `File.m` class, following the issue #282 
The changes are implemented into `File2.m` file, to keep `File.m` as reference.
Corresponding test unit is in file `test_bids_file2.m`

Idea behind the implementation is to let user to set up prefix, entities extension etc manually by setting
corresponding attributes, and automatically update filename, path and json name to retrieve.
Actual implementation is using standard class, but it could be beneficial to move to handle class, to not copy object every time.

Each time the attribute is set, it pass the validation via regexp, which are implemented as static functions, so can be used outside the class.

The return values of `filename`, `json_filename`, `bids_path` are stored in internal field, and updated once user changes entity. It can be beneficial to not store these values, but just calculate each time they are required. To discuss.

To do:
- [x] Implement schema
- [ ] ~~Make it handle class (to discuss)~~ (will see on another PR)
- [x] Integrate into `bids-matlab` error reporting
- [x] replace older implementation
- [x] Documentation
- [x] Update notebook

---

Possible improvements:

- [ ] enforce camel case if user inputs several words for an entity label


Remi-Gau:
Thanks @nbeliy 

Will have a look at this.

tests/test_bids_file2.m
```diff
@@ -0,0 +1,79 @@
...
+  assertEqual(file.filename, 'wuasub-01_ses-test_task-faceRecognition_run-02_bold.nii');
+  assertEqual(file.json_filename, 'wuasub-01_ses-test_task-faceRecognition_run-02_bold.json');
+  assertEqual(file.bids_path, fullfile('sub-01', 'ses-test'));
+
+  filename = [];
+  filename.prefix = 'wua';
+  filename.suffix = 'bold';
+  filename.ext = '.nii';
+  filename.entities = entities;
+  assertEqual(file.filename, 'wuasub-01_ses-test_task-faceRecognition_run-02_bold.nii');
```
Remi-Gau, line 47: I am not clear as to what this is testing. Line 47 is the same as line 38 and the `filename` is not used in this test.

Remi-Gau (changes requested):
First quick pass. This will definitely be a more intuitive API for sure.

### Readability

- [ ] Move constructor to the top of the class definition before the getters and setters

### Style

- [x] use snake_case for all methods name (eg `obj.validateString` --> `obj.validate_string` for consistency

Remi-Gau:
> Actual implementation is using standard class, but it could be beneficial to move to handle class, to not copy object every time.

Should we keep it as a standard class for now, and refactor into a handle class in a separate PR later: this should not affect the actual API, correct?

Remi-Gau:
> Implement schema

Regarding this aspect and a couple of other things that the original File class could do: how do you want to proceed ?

Do you want me to add the tests and let you update the code ?  Do you want to take care of both?

tests/test_bids_file2.m
nbeliy: Line 47 supposed to tests the creation out of structure instead of filename on line 29. But I forgot to initialize file :
```matlab
file = bids.File2(filename);
``` 

Feel need to write tests for tests files

nbeliy:
> Should we keep it as a standard class for now, and refactor into a handle class in a separate PR later: this should not affect the actual API, correct?

The only difference in API is for non-setter functions you will not need to do `file = file.Update();`, a simplier
`file.Update();` will suffice. Same goes with `SetEntity` function.

Another change (more annoying), is for 
[copying classes](https://nl.mathworks.com/help/matlab/matlab_oop/custom-copy-behavior.html),
instead of standard `file2 = file1;` you will need to use `file2 = copy(file1);`.

That's the main reason why I didn't started with handle directly.
I'm not sure that simplification of manual update will justify counter-intuitive for Matlab copy.

nbeliy:
> Regarding this aspect and a couple of other things that the original File class could do: how do you want to proceed ?
> 
> Do you want me to add the tests and let you update the code ? Do you want to take care of both?

For schema, probably you be able to that better - you have better understanding of how it works.

For additional features, can you do the list of required features (already implemented in `File.m` and not yet implemented), and I can incorporate them.

nbeliy:
Fixed readability and naming style.

This auto lint is separate commab or matlab built-in?
I was searching to automatic style corrector that could work without matlab gui

Remi-Gau:
> This auto lint is separate commab or matlab built-in?
> I was searching to automatic style corrector that could work without matlab gui

nope it is the miss_hit python package: 

https://github.com/bids-standard/bids-matlab/blob/master/CONTRIBUTING.md#matlab-code-style-guide-and-quality

You can run this from the command line:
```
mh_style . --fix
```

Or you can install pre-commit hooks defined in the [pre-commit yaml](https://github.com/bids-standard/bids-matlab/blob/dev/.pre-commit-config.yaml) file (`pre-commit install`) to make sure only "linted" things make it into your commit.

Remi-Gau:
> For schema, probably you be able to that better - you have better understanding of how it works.

Okidoki

> For additional features, can you do the list of required features (already implemented in `File.m` and not yet implemented), and I can incorporate them.

I will cross check before making a list because it might be that you might have implemented some already.

Do you want me to take care of the error handling?

Remi-Gau:
> Fixed readability and naming style.

Just checking: did you push the changes?

nbeliy:
> > Fixed readability and naming style.
> 
> Just checking: did you push the changes?

Done, took me some time to understand that instead of push I repeated to pull.
I'll do errors

Remi-Gau:
> Done, took me some time to understand that instead of push I repeated to pull.
> I'll do errors

OK

I work on the schema stuff by doing a pull request on your `nbeliy:crc_improved_File` branch: I would prefer to avoid annoying merge conflicts.

+bids/File2.m
```diff
@@ -0,0 +1,239 @@
...
+
+  methods
+
+    function obj = File2(varargin)
+      args = inputParser;
+      charOrStruct = @(x) isstruct(x) || ischar(x);
+
+      args.addRequired('input_file', charOrStruct);
+      args.addParameter('use_schema', false, @islogical);
+      args.addParameter('tolerant', true, @islogical);
```
Remi-Gau, line 32: Realising that overall bids-matlab is not super consistent in when something should be called as an argument or as "parameter: value" pair.

I think I would prefer to standardize this a bit before the first release. Will open an issue to plan this.

```matlab
BIDS = bids.layout(root, ...
                               use_schema,  ...
                               index_derivatives,  ...
                               tolerant,  ...
                               verbose)

query = bids.query(BIDS, filter)

bids.copy_to_derivative(BIDS, pipeline_name,...
                            out_path, ...
                            filters, ...
                            'unzip', true, ...
                            'force', false, ...
                            'skip_dep', false, ...
                            'use_schema', use_schema, ...
                            'verbose', true);

file = bids.File2(filename, 'use_schema', false ...
                                         'tolerant', true);

report = bids.report(BIDS, ...
                     'output_path', output_path, ...
                     'read_nifti', true, ...
                     'verbose', true);
```

Remi-Gau:
@nbeliy opened a PR on your branch.

Will add a todo item to the top comment of the PR to remember to update the jupyter notebook.

nbeliy:
Integrated errors into `error_handeling`

nbeliy:
There 2 things about errors:

 - matlab don't show error id when printing errors, so instead of `bids:File2:emptySuffix:no suffix specified` user will see only `no suffix specified`, but he will see the stack
 - The switch from tolerant = true to tolerant = false not implemented, that switch need to perform all checks for empty entities and required fields. I'll do it after schema implementation.

nbeliy:
When I open `File2.m` in matlab editor, I see a warning that I can't resolve:
 `l92: a set method for non-Depandent property should not access another property`, in this case it's access `obj.tolerant` and `obj.changed`. In help, they say that matlab do not grantee the order of variables initialization, and this access may lead to variable not found error. 

On my tests, the initialization of variables bypass the set function, so it should not cause any problem, but probably I miss something.

Remi-Gau (approved):

Remi-Gau:
OK this finally goes in: sorry for the delay @nbeliy

PR #300, opened 2022-01-14
In improvements into schema loading/checking

nbeliy:
This merge request re-implements the Schema class , under the name of Schema_imp.
The idea behind is to straightforward schema loading and to add support of different versions of BIDS.
The schemas themselves are stored in folder `schema` as json files:
```
schema/schema_entities_v1.4.1.json
schema/schema_entities_v1.5.0.json
schema/schema_entities_v1.6.0.json
```
These json files was automatically generated from yaml files in rules, and for all tags in BIDS-specifications repository.

The `Schema_imp` load a given version (or last version if passed version is an empty string) into a map with keys
composed out `<datatype>_<suffix>`, and values are same as `suffix_group`, for ex. for func/bold entry it will be:
```
key: funk_bold
value:
    struct with fields:

    extensions: {3x1 cell}
      entities: {10x1 cell}
      required: {2x1 cell}
```

The test of name validity is moved from `append_to_layout` to `Schema_imp.test_name`, which accepts the file name or file structure `p` and the modality.
It returns if file is valid, and `suffix_group` structure.

Other changes: I made the path in `bids.root_dir()` persistant.

Things to do:
- [ ] Documentation (as usual)
- [ ] regexp-related functions and list of suffixes/extentions for given modality
- [ ] tests


Remi-Gau:
I am not too keen on changing the shape of the schema compare to its original yaml version and on dropping the rest of its content.

Also is there any urgent need to support several versions of the schema? 

I know that this is often requested on the bids validator side specifically to check that a given dataset is valid (given the BIDS version in its dataset_description) against the corresponding version of the schema, but I am not sure this is needed for bids matlab as far as I can see.

nbeliy:
> I am not too keen on changing the shape of the schema compare to its original yaml version and on dropping the rest of its content.

In theory agree, but (a) the form of original yaml changes from version to version, (b) matlab don't officially support yaml.
As I understand, in specifications rules are used to generate different tables, so they organized to facilitate such generation.
For bids interfaces we need only the list of entities, and list of required ones, a lot of info in official rules are not needed.
But I will not fight over the exact structure of json.

 > Also is there any urgent need to support several versions of the schema?

Urgent no, but it's relatively easy to implement.
Imagine a dataset made following 1.6.0, and a pipeline based on bids-matlab to treat it.
Then in half of year the version 2.8 comes out, bids-matlab will be updated, and pipeline will become broken --
previously conform dataset will become not conform anymore. 
For sure it will be easy to fix (just changing the tolerant value), but still annoying.
Giving support of different versions will avoid that problem.

In grander picture, having a maintained set of rules for all versions (possibly in separate repository) will help
developers (eg. me) of bidsifiers/validators to avoid a lot of if/else when dealing with older bids datasets.
Hoping introduce it into bids-matlab and bidsme, then go to bids with user-case.

Remi-Gau:
> In theory agree, but (a) the form of original yaml changes from version to version, 

yeah sort of hoping that things will stabilize and will try to keep the code fairly agile to adapt to new changes if needed.

> (b) matlab don't officially support yaml. 

yeah I know. Had to set up a github action to automate the conversion and deliver a json version every time there is a change upstream

https://github.com/bids-standard/bids-matlab/blob/master/.github/workflows/update_schema.yml

> As I understand, in specifications rules are used to generate different tables, so they organized to facilitate such generation. For bids interfaces we need only the list of entities, and list of required ones, a lot of info in official rules are not needed. 

Yeah it does add a bit of clutter but I think we can live with this for now.


> Urgent no, but it's relatively easy to implement. Imagine a dataset made following 1.6.0, and a pipeline based on bids-matlab to treat it. Then in half of year the version 2.8 comes out, bids-matlab will be updated, and pipeline will become broken -- previously conform dataset will become not conform anymore. For sure it will be easy to fix (just changing the tolerant value), but still annoying. Giving support of different versions will avoid that problem.

Couple of reasons why I am not too worried:
- a move to BIDS 2.0 is not going to happen any time soon, and I would be surprised if stable schema version did not happen before. So breaking backward compatibility should not be an issue for quite a while.

> In grander picture, having a maintained set of rules for all versions (possibly in separate repository) will help developers (eg. me) of bidsifiers/validators to avoid a lot of if/else when dealing with older bids datasets. Hoping introduce it into bids-matlab and bidsme, then go to bids with user-case.

I agree that this is an important issue but this goes beyond BIDS matlab: will open a discussion on the main BIDS repo because I doubt you are the only one to worry about this.

Remi-Gau:
Sounds good me.

Issue #470, opened 2022-11-08
[BUG] Empty file index during loading layout  qMRI reference dataset


nbeliy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating system

- [X] Linux
- [ ] Mac
- [ ] Windows

### Operating system version

```Markdown
- Linux Ubuntu 18.04 "bionic beaver"
```

### Platform

- [X] MATLAB
- [ ] Octave

### Platform version

```Markdown
- MATLAB 2017a
```

### bids-matlab version

v0.1.0

### branch / commit number

```Markdown
- main 913246915448505b43c
```

### Current Behavior

While loading layout the [ds-mpm](https://osf.io/k4bs5/) from OSF, the `bids.layout` crashes with `Expected one output from a curly brace or dot indexing expression, but there were 0 results.` error.

From quick invistigation, problem comes from file `sub-01/fmap/sub-01_acq-headMTw_RB1COR.json` during searching dependencies.

The returned `file_info` structure (`layout.m:597) has an empty file_idx:
```matlab
info_dest = 

  struct with fields:

        path: '/Storage/beliy/bidsifications/osf/source/sub-01/fmap'
    filename: 'sub-01_acq-headMTw_RB1COR.json'
     sub_idx: 1
    modality: 'fmap'
    file_idx: [0x1 double]
```

### Error message

```Markdown
>> BIDS = bids.layout('source')
Expected one output from a curly brace or dot indexing expression, but there were 0 results.

Error in bids.layout>manage_dependencies (line 598)
      BIDS.subjects(info_dest.sub_idx).(info_dest.modality)(info_dest.file_idx) ...

Error in bids.layout (line 148)
  BIDS = manage_dependencies(BIDS, verbose);
```

### Anything else?

I will give more details later, so far the bug is quite mistirious.

Remi-Gau:
I can reproduce it on the dev branch.

Also getting this error for 'ds-vfa'.

```
bids.layout(fullfile(pwd, 'ds-vfa'))
Dot indexing is not supported for variables of this type.

Error in bids.internal.return_file_index (line 15)
  file_idx = strcmp(filename, {BIDS.subjects(sub_idx).(modality).filename}');

Error in bids.internal.return_file_info (line 20)
  file_info.file_idx = bids.internal.return_file_index(BIDS, ...

Error in bids.layout>manage_dependencies (line 663)
      info_dest = bids.internal.return_file_info(BIDS, dest);

[...]

PR #473, opened 2022-11-09
[FIX] cleaner handling of missing dependency

nbeliy:
### All Submissions:

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/bids-standard/bids-matlab/pulls) for the same update/change?

Changes: added a test if file referenced by IntendedFor field is indexed (if `isempty(info_dest.file_idx)`).
Now it produces warning (if verbose) and skips the file:
```
Warning: IntendedFor file /Storage/beliy/bidsifications/osf/source/sub-01/fmap/sub-01_acq-headMTw_RB1COR.json from sub-01_magnitude1.nii not indexed
```

Removed the try/catch introduced in #471 as it may hide possible bugs and blame input datasets.

This try/catch can be re-introduced, but I feel it must protect full `manage_dependencies` function.

Making request draft, as there an alternative implementation of the fix #474.

### DIFF
--- a/+bids/layout.m
+++ b/+bids/layout.m
@@ -598,7 +598,7 @@ function BIDS = manage_dependencies(BIDS, verbose)
       if isempty(info_dest.file_idx)
         msg = ['IntendedFor file ' dest ' from ' file.filename ' not indexed'];
         bids.internal.error_handling(mfilename, 'IntendedForMissing', msg, tolerant, verbose);
-        continue;
+        continue
       end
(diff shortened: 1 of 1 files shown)


Remi-Gau:
closes #470 

I have a preference for this one if only because all the tests pass out of the box.

Remi-Gau (approved):
LGTM.

Thanks @nbeliy

Issue #476, opened 2022-11-10
Excessive warnings when generating derivatives under windows


nbeliy:
Not sure if it is intended or a oversight (so deviating from template).

commit: master df2aa00955

In `+bids/copy_to_derivative.m:307
```matlab
if  isunix
   command = sprintf('cp -R -L -f %s %s', src, target);
   status = system(command); 
elseif ispc
  msg = 'Unknown system: copy may fail';
  bids.internal.error_handling(mfilename, 'copyError', msg, true, verbose);
  use_copyfile(src, target, unzip_files, verbose);
else
    use_copyfile(src, target, unzip_files, verbose);
end
```
The warning `Unknown system` happens in `ispc` section, while logically is should be in `else` part.
Is this designed like that (and then is little bit annoying with warning trace on)?

If needed I can make an PR.

Remi-Gau:
oh indeed.
not sure why I am not seen that when testing on windows.

in any case, yes open a PR please. much appreciated. 👍🏾

Remi-Gau:
actually some that copying logic is to more safely handle datalad datasets: after your PR I will probably add something that makes it more explicit.

nbeliy:
Made the PR #485

PR #485, opened 2022-11-14
Fixes issue #476

nbeliy:
Addresses the issue #476 with excessive warnings during the copy to derivative under windows.

Moved warnings in `bids.copy_to_derivatice.m:307` from `if ispc` block to `else` block


Remi-Gau (approved):
LGTM

Thanks @nbeliy

PR #532, opened 2023-03-28
[FIX] transformers with "filter" should not fail when filtering variable is missing

Remi-Gau:


Remi-Gau:
@all-contributors add @avanaudenhaege for bug

PR #534, opened 2023-03-29
[INFRA] Use PAT to submit pull requests

effigies:
This will ensure that actions run on the PR. The default `GITHUB_TOKEN` is not enabled to do that to avoid a possible loop of new PRs with actions.

### All Submissions:

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/bids-standard/bids-matlab/pulls) for the same update/change?

PR #537, opened 2023-04-11
[BOT] update schema

effigies:
done via this [GitHub Action](https://github.com/bids-standard/bids-matlab/blob/master/.github/workflows/update_schema.yml)


Remi-Gau:
@effigies 
since #534 I am getting these weird PRs (see previous case https://github.com/bids-standard/bids-matlab/pull/535)

Before I start digging, if you have a hunch on what could now be causing this (a hunch of the the type: I have seen this before), let me know.

PR #590, opened 2023-07-10
[ENH] index and query phenotype

### LINKED ISSUE #589: index content of phenotype folder

Remi-Gau:
closes #589

### DIFF
--- a/+bids/layout.m
+++ b/+bids/layout.m
@@ -116,6 +116,7 @@ function BIDS = layout(varargin)
   % BIDS.description   -- content of dataset_description.json
   % BIDS.sessions      -- cellstr of sessions
   % BIDS.participants  -- for participants.tsv
+  % BIDS.phenotype     -- for content of the phenotype folder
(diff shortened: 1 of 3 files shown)


Remi-Gau:
I would prefer to have:

- speed: not loading any tsv, or json is faster
- consistency: either load all json and tsv or none of them
- ease of use: making it easier for users to access things via query, rather than query + load file content 

For now we have

- for scans.tsv and sessions.tsv are not loaded when running bids.layout and actually I think that querying them is still a bit of a mess.
- events.tsv and things in the datatype folders are not loaded
- Participants.tsv is loaded.
- Phenotype/*.tsv... I am not sure: usually you start seeing those files in large datasets with many participants and they can be fairly large files, though I doubt this will ever be the bottleneck compared to the time it just takes to indexing a lot of files.

Remi-Gau:
For consistency the content of phenotype tsv / json will not be loaded

Issue #369, opened 2022-05-10
`bids.File.rename` should rename the json sidecar too 


Remi-Gau:
rename method for bids.File should also rename the json  side car and most likely transfer all metadata when doing that

add a method to update metadata?

PR #597, opened 2023-07-28
Incorporating json into bids.File (issues #596 #371)

### LINKED ISSUE #371: Add a method to `bids.File` for changing file metadata
> Add a function for changing one or several fields in a .json metadata file (e.g., to generate a new metadata file after applying some transformation / preproc step to a dataset)

_Originally posted by @Remi-Gau in https://github.com/bids-standard/bids-matlab/issues/79#issuecomment-1123227853_

### LINKED ISSUE #596: [ENH]: Incorporate json manipulation into bids.File class
### Is there an existing issue for this?

- [X] I have searched the existing issues

### New feature

```Markdown
Playing with `bids-matlab`, I found that `bids.File` class is extremely useful even outside working with bids layouts, as it allows to quickly create/modify the filenames, especially useful when creating derivatives.

It would be nice that the same class could modify the sidecar json file, making easier to supplement derivative with it's json.

As an example, by using `spm_reorient`:
```matlab
% out_list is the list of reoriented images
for iFile = 1:numel(out_list)
  p = bids.File(out_list{iFile});
  p.load_json(); % Loads the existing sidecar json, or create empty structure
  p.entities.space = space; % Adding space entity

  p.update_json('Description', ['Reoriented to ' space], ...
                'Sources', out_list{iFile},...
                'SpatialReference', ref_img{1});

  outfile = fullfile(pth, p.filename);
  movefile(reoriented, outfile); % Renaming the reoriented file
  p.export_json(pth); % Saving modified json to the path
end
```

I already have these functions as standalone, integrating them into `bids.File` shouldn't take a lot of time. 

If this interests you, I can create a PR.
```


### Unclear documentation

_No response_

### PR DESCRIPTION
**Pull request by nbeliy**
First draft to incorporating managing sidecar into `bids.File` class.

@Remi-Gau , Is it better to use the `bids.File` class directly or use a daughter class extending functionality?

Todo:
 - [x] lint
 - [x] tests
 - [x] docs
 - [x] Decide if create new class `bids.JSONFile`or use the `bids.File`
 - [ ] Add path management to `Sources` and `Raw`. BIDS requires path starting from root of dataset

---

- closes #596 
- Closes #371

### CODE REVIEW

**Comment by Remi-Gau** [comment]
> Is it better to use the `bids.File` class directly or use a daughter class extending functionality?

my hunch would be to first keep things in the same class

**Comment by nbeliy** [comment]
Ok, will merge into main class.

**Comment by nbeliy** [comment]
Few changes:
  - Removed JSONFile subclass, and moved functions to File class
  - Created new `util.update_struct` function that manages the updating fields in json structure (with tests)
  - Added documentation

@Remi-Gau 
  - `util.update_struct` duplicates the internal function from `bids.Description`, should we replace it?
  - In current `bids.File` implementation, the metadata is loaded when instance is created, which may cause slowdowns (reading file on disk and searching for inheritence). Will it be better to make metadata load on request only?
  - I've added the doc for both `util.update_struct` and `File`, can you check if it's extensive and comprehensive?

File: +bids/+util/update_struct.m [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+  %      a: 'new_val'
+  %      b: 'val_b'
+  % Removing field from structure:
+  %   update_struct(struct('a', 'val_a', 'b', 'val_b'),
+  %                 'b', [])
+  %   struct with fields:
+  %      a: 'val_a'
+  % Appending values to existing field:
+  %   update_struct(struct('a', 'val_a', 'b', 'val_b'),
+  %                 'b-add', 'val_b2')
```
Remi-Gau, line 31:
```suggestion
  %                 'b-add', 'val_b2')
```
Random thought.
I know this is an edge case in the world of bids because we would expect the field to be camel camse but what if one of the field to update is named "b-add", do you have then do something like: "b-add-add" ?

Remi-Gau: maybe it would be better to have simpler function:
- one for adding
- one for removing
- one for appending an element to a an array or a cell
- one for popping an element to an array or a cell

or have the first argument specify what "action" to perform.

nbeliy: Well, no matter how BIDS will decide to make life needlessly complicated, but if json is transformed into matlab structure, the fieldnames like `b-add` will be forbidden.

Splitting into several subcommand: `metadata_add`, `metadata-append`, `metadata-rm` can be implemented, and maybe is a good idea. But then it would be difficult to work with struct instead of parameters.

For poping (remove and return) is more complicated, as matlab do not modify current object, and multiple returns looks ugly:
```matlab
 [file, val] = file.metadata_pop('val_a');
```

What is your take on not forcing to read metadata when creating File instance, and making it on request only?

Remi-Gau: > the fieldnames like `b-add` will be forbidden

Oh dear... Matlab 101. Silly me for forgetting this... I guess that's what happens when you start switching back and forth between languages

nbeliy: I agree that inheritance should be part of bids (even if I don't like it due to all the complications). My question should the metadata be loaded implicitly or explicitly (`f.set_metadata();`). 

Implicit reading of metadata imposes that the logic of retrieving json files, and reading them happens every time when a new instance is created, even if it's not needed. 

The way how inheritance principle is implemented is an another issue -- it requires the knowledge of the layout of dataset, which is stored in layout structure, but not in File. For me (very personal opinion), inheritance principle should be applied during creation of layout, by setting File.metadata_files, but not in File.

nbeliy:
Added tests for File and set up implicit metadada reading.

For me, the PR can be removed from draft.

+bids/+util/update_struct.m
Remi-Gau: Just realized that `set_metadata` relies on `bids.internal.get_meta_list` that tries to implement the inheritance principle (same one use for bids.layout).

```
    function obj = set_metadata(obj)
      if isempty(obj.metadata_files)
        pattern = '^.*%s\\.json$';
        obj.metadata_files = bids.internal.get_meta_list(obj.path, pattern);
      end
      obj.metadata =  bids.internal.get_metadata(obj.metadata_files);
    end
```

I would say that changing this is beyond the scope of this PR which is on updating metadata. Just thought it was good to mention it.

Remi-Gau: > For poping (remove and return) is more complicated, as matlab do not modify current object, and multiple returns looks ugly:
> 
> ```matlab
>  [file, val] = file.metadata_pop('val_a');
> ```

I think I would just return the file object.

```
file = file.metadata_pop('FieldToPopFrom', 'value_to_pop');
```

+bids/File.m
```diff
@@ -707,6 +739,48 @@ function check_required_entities(obj)
...
+      % USAGE::
+      %
+      %  f = f.metadata_update(key1, value1, key2, value2);
+      %  f = f.metadata_update(struct(key1, value1, ...
+      %                        key2, value2));
+
+      obj.metadata = bids.util.update_struct(obj.metadata, varargin{:});
+    end
+
+    function out_file = metadata_export(obj, varargin)
```
Remi-Gau, line 760:
```suggestion
    function out_file = metadata_write(obj, varargin)
```

Maybe write to stay consistent with tsvwrite? Or `encode` to stay consistent with `jsonencode`? WDYT?

nbeliy:
> Tacking https://github.com/bids-standard/bids-matlab/issues/369 feels a bit too much in this PR, no?
maybe better in a separate one, no?

Technically yes, but it can be integrated in the current PR. But I'm not sure what to do with inheritance principle -- renaming can easily brake it.

Remi-Gau:
You are right. One headache at time. Let's leave it out of the scope of this one for now.

+bids/+util/update_struct.m
Remi-Gau: FWI: will open a small PR to add a parameter to `bids.internal.get_meta_list` to decide to use the inheritance principle or not to list metadata files.

It could then be reused here to let user decide what they want to do.

nbeliy:
1. Added functions that acts on single field:
  - `metadata_add(obj, field, value)` to add/overwrite -- pointless, as same effect can be achieved with `f.metadata.Filed = value`
  - `metadata_append(obj, field, value)` -- transforms value to cellarray if needed, and appends new value
  - `metadata_remove(obj, field)` -- removes field from metadata, don't want to call it pop, cause in python remove removes element, and pop removes and returns element

The `metada_update`  is kept, as a way to update several fields at same time without copying same variable 25 times (why matlab don't have pointers or references????)

Updated doc and tests (couldn't run coverage).

At this point the PR do what is supposed to do -- ready for final discussion/review?

I'll open a  new tangent issue for automatic loading of metadata, just to keep things organized.

+bids/File.m
```diff
@@ -65,6 +65,65 @@
   %
   %   file = bids.File(name_spec, 'use_schema', true);
   %
+  % Load metadata (supporting inheritance).
+  %
+  % .. code-block:: matlab
+  %
+  %   f = bids.File('tests/data/synthetic/sub-01/anat/sub-01_T1w.nii.gz');
+  %
```
Remi-Gau, line 73: need to check but maybe better to put the help section after each method than all of them at the top

if the file has all the method help at the top at the moment, then let's take care of that in a separate PR.

Remi-Gau (approved):
2  minor points regarding having smaller tests and where the doc for each method should go.

If you don't have the time I can take care of it.

just let me know.

thanks

tests/test_bids_file.m
```diff
@@ -67,6 +67,50 @@ function test_get_metadata_suffixes_basic()
...
+  % metadata_update
+  bf = bf.metadata_update('Testing', 'adding field');
+  assertTrue(isfield(bf.metadata, 'Testing'));
+  assertEqual(bf.metadata.Testing, 'adding field');
+
+  bf = bf.metadata_update('Testing', 'modifying field');
+  assertEqual(bf.metadata.Testing, 'modifying field');
+
+  bf = bf.metadata_update('Testing', []);
+  assertFalse(isfield(bf.metadata, 'Testing'));
```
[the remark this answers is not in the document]
nbeliy: Can do -- I put there to not reread metadata again.

+bids/File.m
nbeliy: It's more an example, making aware that such functions exists. The actual doc are already in the functions below.

I think it's worth to keep the doc as it is -- long descriptions are common in matlab.

tests/test_bids_file.m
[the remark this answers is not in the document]
Remi-Gau: yeah that's always the trade off, but unless set up is super heavy then I tend to prefer smaller independent tests.

+bids/File.m
Remi-Gau: OK that makes sense. good with me

tests/test_bids_file.m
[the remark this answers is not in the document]
nbeliy: Just to be clear, do you want a new function inside `tests_bids_file.m` or completely separate test suite? I guess it's former, but want to be sure.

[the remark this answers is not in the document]
Remi-Gau: > a new function inside `tests_bids_file.m`

just new function

matlab is annoying enough to force you to have one "main" function per file, let's not feed the monster. 😉

nbeliy:
Spitted the tests.

Just to be completely transparent and to be sure that docs are clear:
`metadata_update`: several keys can be passed, or one structure
`metadata_add`, `metadata_append`, `metadata_remove`: only one key can be passed; do not catches errors
`metadata_write`: passed keys modifies only written metadata, not the original object

Remi-Gau:
All good.
thanks for this @nbeliy 
looking forward to start using this actually!!

Remi-Gau:
I may add another option later in a separate PR to add the possibility to use the inheritance principle or not when loading the metadata 

will ask for a review if you have time

PR #656, opened 2023-11-21
Added modality retrieval from path

### LINKED ISSUE #638: [ENH]: Retrieval of modality by bids.File
### Is there an existing issue for this?

- [X] I have searched the existing issues

### New feature

So far, the modality is the only field which is not retrieved by `bids.File`, which can cause difficulties in file names manipulations, for ex. when I want to cite a source file in derived one:

```
p = bids.File(source_file);
sources = {fullfile(p.bids_path, p.filename)}

sources =

  1x1 cell array

    {'sub-KTA011/sub-KTA011_recording-manual_blood.csv'}
```

the modality is missing and must be set manually.

As the way to retrieve  modality, it can be done by analyzing `p.path`:
```
[path, ~, ~] = fileparts(p.path);
[path, modality, ~] = fileparts(path); % Modality is just one level above the file
[path, entity, ~] = fileparts(path); % entity (sub or ses) is just above the modality
```
We can just check if `entity` matches with `ses` (or `sub`, if there no session), and if it is, then modality is found. If not, the file in question is not inside the BIDS dataset.
```

nbeliy:
- closes #638

---

- based on issue  #638 added a function `get_modality` for `bids.File`.

Modality retrieval is quite simple:

1. From path get the latest directory as modality candidate
2. Compare `ses` (or `sub`, if no sessions) entity with directory above, if it's matches, then modality candidate is accepted.

If concept is approved, I'll update the tests.


Remi-Gau:
sorry it took me this long to reply

Yes that makes sense to me so far: I will let you add / fix tests. As usual, I suspect this will be fairly robust when working with vanilla BIDS but could get funky when not working on validated BIDS dataset but so be it.

PR #767, opened 2025-08-16
[DOC] restructure documentation

Remi-Gau:
- move documentation on how to run tests and build doc in the actual documentation
- fix typos, links...
- add github action to use sphinx to check the links


.github/workflows/validate.yml
```diff
@@ -21,13 +21,3 @@ jobs:
...
-
-  markdown_link_check:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-    - uses: gaurav-nelson/github-action-markdown-link-check@v1
-      with:
-        use-quiet-mode: yes
-        use-verbose-mode: yes
-        config-file: .github/workflows/mlc_config.json
```
Remi-Gau, line 33: removing because we can just rely on sphinx to check the links in the doc

PR #779, opened 2025-08-28
[MAINT] Several changes

markmikkelsen:


.gitignore
```diff
@@ -77,3 +77,10 @@ octave-workspace
 # javascript
 node_modules/*
 yarn.lock
+
+# tests
+dummy_ds/
+
+# node/npm
+package.json
```
Remi-Gau, line 85:
```suggestion
```
we already have it in the repo, so I think it's fine to keep it

Remi-Gau:
ok tests are failing on main too so it's not this PR
let me try to fix main first

+bids/+util/download_ds.m
```diff
@@ -86,7 +86,7 @@
   end
 
   % clean previous runs
-  if isdir(out_path)
+  if isfolder(out_path)
```
Remi-Gau, line 89: I think I was using isdir because older versions of octave did not have isfolder.

Seems that Octave >= 5 has it:
https://octave.org/NEWS-5.1.html#alphabetical-list-of-new-functions-added-in-octave-5

Since the Matlab linter is commplaining about this, we could replace all `isdir` in a separate PR.

markmikkelsen: Sounds good, I can do that.

PR #788, opened 2025-09-09
[FIX] Clarify error message for invalid label characters

markmikkelsen:
Improves the error message when a label contains invalid characters, specifying that only alphanumeric and "+" are allowed.


Remi-Gau:
tests failures seem unrelated: mostly some flaky tests

PR #805, opened 2026-02-20
FIX: copy_to_derivative.m adds duplicate entries in dataset_description.json

marcelzwiers:
Every time copy_to_derivative.m is called, it appends the same `GeneratedBy` item to the dataset_description file. This commit checks for that.


+bids/copy_to_derivative.m
```diff
@@ -141,20 +141,17 @@ function copy_to_derivative(varargin)
...
-  if exist(descr_file, 'file')
+  if isfile(descr_file)
     content = bids.util.jsondecode(descr_file);
     ds_desc = ds_desc.set_field(content);
-    ds_desc = ds_desc.append('GeneratedBy', struct('Name', args.Results.pipeline_name));
-
-  else
-    ds_desc = bids.Description(args.Results.pipeline_name, BIDS);
-
+    if ~contains(jsonencode(content.GeneratedBy), ...
```
Remi-Gau: cannot use `contains` with octave

+bids/copy_to_derivative.m
```diff
@@ -141,19 +141,24 @@ function copy_to_derivative(varargin)
     bids.util.mkdir(derivatives_folder);
   end
 
+  % update dataset_description
+  % but only if the current pipeline is different
+  % from the one stored in the last entry
+  % of the GeneratedBy section
```
Remi-Gau, line 147: @marcelzwiers 
slightly edited your fix to only check the last entry of the generated by

Remi-Gau:
still some tests failing only on windows: I doubt this is related to this the change in the code. I suspect we need to do some post test tear down.

